PR #2, opened 2017-05-08
Importing more commons from fgbio (and a little tidying)

tfenne:


nh13 (approved):
We can improve test coverage later, but for our purposes, lgtm.

PR #8, opened 2017-05-12
Test methods for capturing stderr, stdout, and Logger.

nh13:
Test coverage went up by 1.5% too!


nh13:
@tfenne you had a suggestion in-person, but I can't seem to recall, can you re-review and see if you remember?

tfenne:
I'd suggested that we synchronize the various methods that take a `runnable: () => Unit` so that if any tests are run in parallel the correct thing continues to happen.

nh13:
@tfenne one more look?

tfenne (approved):

PR #9, opened 2017-05-12
Added a self-closing iterator class.

tfenne:


src/main/scala/com/fulcrumgenomics/commons/collection/SelfClosingIterator.scala
```diff
@@ -0,0 +1,55 @@
...
+    */
+  override def next(): A = {
+    val result = super.next()
+    if (!hasNext) close()
+    result
+  }
+
+  /** Invokes the closer function provided in the constructor. */
+  override def close(): Unit = if (open) {
+    open = true
```
nh13: Uh, shouldn't this be `open = false`?  Could you add a test that you check fails before you change it?

src/test/scala/com/fulcrumgenomics/commons/collection/SelfClosingIteratorTest.scala
```diff
@@ -0,0 +1,30 @@
+package com.fulcrumgenomics.commons.collection
+
+import com.fulcrumgenomics.commons.util.UnitSpec
+
+class SelfClosingIteratorTest extends UnitSpec {
+  "SelfClosingIterator" should "return all the items and then call the close function" in {
```
nh13, line 6: This test wont check that `() => closed` isn't called twice.  A better test would be to have
`var closed = 0` and `() => closed += 1`, and then check the value of `closed` (should be `1`).

nh13 (changes requested):
A little but important bug to fix.

src/test/scala/com/fulcrumgenomics/commons/collection/SelfClosingIteratorTest.scala
tfenne: Added another test.

src/main/scala/com/fulcrumgenomics/commons/collection/SelfClosingIterator.scala
tfenne: Doh.  I changed the var from `closed` to `open` and apparently missed this.  Thanks for the catch.

tfenne:
All fixed up and two more tests.  Want to take a second look @nh13?

nh13 (approved):
LGTM

PR #10, opened 2017-05-20
Update the version to include a git-hash.

nh13:
Also add a task, 'shadowJar', that creates a JAR without the trailing
<git-hash>-SNAPSHOT.jar.


nh13:
@tfenne bump.  I want to do this for all our repos if you are OK with it.

tfenne (approved):
I'm not sure I totally follow the changes, but let's give it a shot!

PR #11, opened 2017-06-01
Bug fix for 2.11.11 builds

nh13:
headOption is not found in BetterBufferedIterator for 2.11.11, but is
found for 2.12.2

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/commons/collection/BetterBufferedIterator.scala
+++ b/src/main/scala/com/fulcrumgenomics/commons/collection/BetterBufferedIterator.scala
@@ -26,11 +26,17 @@ package com.fulcrumgenomics.commons.collection
 
 import com.fulcrumgenomics.commons.CommonsDef._
 
+/** A little trait to allow BetterBufferedIterator to override the headOption method. The latter method
+  * is not found in [[BetterBufferedIterator]] in 2.11.11, but in 2.12.2! */
+sealed trait HeadOption[A] {
+  def headOption: Option[A]
+}
+
 /**
   * A better buffered iterator that provides implementations of takeWhile and dropWhile
   * that don't discard extra values.
   */
-class BetterBufferedIterator[A](private val iterator: Iterator[A]) extends BufferedIterator[A] {
+class BetterBufferedIterator[A](private val iterator: Iterator[A]) extends HeadOption[A] with BufferedIterator[A] {
   private var buffer: Option[A] = maybeNext
 
   /** Returns a Some(A) if there is a next in the iterator else None. */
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/commons/collection/BetterBufferedIterator.scala
```diff
@@ -26,11 +26,17 @@ package com.fulcrumgenomics.commons.collection
 
 import com.fulcrumgenomics.commons.CommonsDef._
 
+/** A little trait to allow BetterBufferedIterator to override the headOption method. The latter method
+  * is not found in [[BetterBufferedIterator]] in 2.11.11, but in 2.12.2! */
+sealed trait HeadOption[A] {
```
tfenne, line 31: I don't think that this trait is needed at all since either clients will know the type to be BetterBufferedIterator and so `headOption` will exist, or it won't know the type beyond "BufferedIterator" in which case it can't call headOption.  I don't think adding the trait gives you anything beyond allowing you to add `override` below.

src/main/scala/com/fulcrumgenomics/commons/collection/BetterBufferedIterator.scala
```diff
@@ -26,11 +26,17 @@ package com.fulcrumgenomics.commons.collection
...
+  def headOption: Option[A]
+}
+
 /**
   * A better buffered iterator that provides implementations of takeWhile and dropWhile
   * that don't discard extra values.
   */
-class BetterBufferedIterator[A](private val iterator: Iterator[A]) extends BufferedIterator[A] {
+class BetterBufferedIterator[A](private val iterator: Iterator[A]) extends HeadOption[A] with scala.collection.BufferedIterator[A] {
```
tfenne: Why the change to using the FQN?

tfenne (commented):
Couple of minor comments.

tfenne:
Oh .... also maybe a trivial test that uses `headOption` to confirm it works the same under 2.11 and 2.12?

src/main/scala/com/fulcrumgenomics/commons/collection/BetterBufferedIterator.scala
nh13: Since I need to use the `override` keyword when `headOption` is found in `BufferedIterator` in 2.12.

src/main/scala/com/fulcrumgenomics/commons/collection/BetterBufferedIterator.scala
nh13: Good catch

src/main/scala/com/fulcrumgenomics/commons/collection/BetterBufferedIterator.scala
```diff
@@ -59,6 +65,13 @@ class BetterBufferedIterator[A](private val iterator: Iterator[A]) extends Buffe
   /** Drops items while they match the predicate. */
   override def dropWhile(p: (A) => Boolean): this.type = {
     while (hasNext && p(head)) next()
+    this.headOption
```
nh13: removed!

PR #14, opened 2017-07-01
Squash a file open leak

nh13:
This gets nasty when reading a lot of files!


tfenne (approved):
Doh.  I thought for sure `Source` called it's close method when it hit the end of it's input.  I guess not.

nh13:
@tfenne it does close it, but I found that it doesn't close it until we have reached the end of the iterator, which makes a lot more sense.  The issue is since this is lazy, if we call `getLines` a lot of files (think all the logs from a big `dagr` pipeline), then we may have too many file handles.  I instead add some more docs warning that the underlying file handle is not closed until the end of the iterator is reached.

PR #19, opened 2017-11-10
Add a method to read lines from a resource on the classpath

nh13:


src/main/scala/com/fulcrumgenomics/commons/io/Io.scala
```diff
@@ -179,4 +179,12 @@ trait IoUtil {
 
   /** Creates an object that will asynchronously read character data from a stream and pipe it into a sink function. */
   def pipeStream(stream: InputStream, sink: String => Unit) : AsyncStreamSink = new AsyncStreamSink(stream, sink)
+
+  /** Finds a resource with a given name on the classpath and produces an iterator of lines of text from the resource. */
+  def readLinesFromResource(name: String): Iterator[String] = {
+    val in = getClass.getResourceAsStream(name)
+    if (in == null) throw new IllegalArgumentException(s"Resource does not exist at path: $name")
```
tfenne: `require( in != null, ....)`?

tfenne (approved):
One minor comment, otherwise looks good.

PR #20, opened 2017-11-29
Add a class to access a type safe configuration object.

nh13:
Extend the trait to provide access to a configuration object of your
choice.  The methods support a larger set of possible types that
can be returned.

NB: this is moving over configuration from https://github.com/fulcrumgenomics/dagr/ into https://github.com/fulcrumgenomics/commons/

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -115,7 +115,8 @@ lazy val root = Project(id="commons", base=file("."))
   .settings(description := "Scala commons for Fulcrum Genomics.")
   .settings(
     libraryDependencies ++= Seq(
-      "org.scala-lang" %  "scala-reflect" %  scalaVersion.value,
+      "com.typesafe"   %  "config"        % "1.3.2",
+      "org.scala-lang" %  "scala-reflect" % scalaVersion.value,
       //---------- Test libraries -------------------//
       "org.scalatest"  %% "scalatest"     % "3.0.1"  % "test->*" excludeAll ExclusionRule(organization="org.junit", name="junit")
     )
--- a/src/main/scala/com/fulcrumgenomics/commons/util/ConfigurationLike.scala
+++ b/src/main/scala/com/fulcrumgenomics/commons/util/ConfigurationLike.scala
@@ -1,0 +1,235 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2017-2018 Fulcrum Genomics LLC
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package com.fulcrumgenomics.commons.util
+
+import java.nio.file.Path
+import java.time.Duration
+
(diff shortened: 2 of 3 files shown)


nh13:
@tfenne bump

src/main/scala/com/fulcrumgenomics/commons/util/ConfigurationLike.scala
```diff
@@ -0,0 +1,247 @@
...
+    * does not exist, an exception is thrown. If the requested type is not supported, an
+    * exception is thrown.
+    */
+  def apply[T : TypeTag](path: String) : T = {
+    this._requestedKeys.add(path)
+    try {
+      asType[T](path)
+    }
+    catch {
+      case ex : Exception =>
```
tfenne: A couple of things here:

1. I think exceptions should _either_ be logged or re-thrown but not both as that will almost always lead to double-logging which can be confusing.  I tend to think that low-level APIs like this should _never_ log and only communicate by succeeding or throwing exceptions.
2. Similar problem with embedding the prior exception's message and also chaining it.  It will lead to some part of the information being printed twice which is more confusing than helpful.  I would get rid of the `: ${ex.getMessage}` and just chain it.

src/main/scala/com/fulcrumgenomics/commons/util/ConfigurationLike.scala
```diff
@@ -0,0 +1,247 @@
...
+  *   }
+  *   scala> conf.getOrElse[String]("a.does-not-exist", "default")
+  *   res0: String = default
+  *   scala> conf.getOrElse[Int]("a.does-not-exist", 0)
+  *   res1: Int = 0
+  *   scala> conf.getOrElse[Set[Int]]("a.does-not-exist", Set.empty)
+  *   res2: Set[Int] = Set()
+  * }}}
+  */
+trait ConfigurationLike {
```
tfenne: This is lifted over from dagr, but with quite a few changes.  I'm really not sure about the move from `def configure()` to `apply()/get()`.  The intent of the trait in dagr is that you could mix it into just about any class to gain access to configuration and there it felt natural to write things like:

```scala
private val limit = configure[Int]("some_limit")
```

Changing it to use `apply()` means that any class that wants to mix in the trait suddenly has a public apply() method that accesses configuration.  Which makes it hard to the class to provide it's own apply() method(s) that suit it's own purposes and also makes for a confusing public API. 

It feels like the intent here is very different - to make it easier to generate `Config` singleton objects?  In which case maybe it should not be a trait but instead be an single class that can be instantiated (e.g. into a case object) or extended to add more functionality?

src/main/scala/com/fulcrumgenomics/commons/util/ConfigurationLike.scala
```diff
@@ -0,0 +1,247 @@
...
+    */
+  def get[T : TypeTag](path: String) : Option[T] = {
+    this._requestedKeys.add(path)
+    if (config.hasPath(path)) Some(apply[T](path))
+    else None
+  }
+
+  /**
+    * Optionally accesses a configuration value. If the value is not present in the configuration
+    * a None will be returned, else a Some(T) of the appropriate type.
```
tfenne: `a None` -> `the default value`

src/main/scala/com/fulcrumgenomics/commons/util/ConfigurationLike.scala
```diff
@@ -0,0 +1,247 @@
...
+  */
+trait ConfigurationLike {
+  /** The configuration to query. */
+  protected def config : Config
+
+  /** A logger to which to report exceptions. */
+  protected def _logger: Option[Logger] = None
+
+  /** The keys requested, in order. */
+  private val _requestedKeys = new java.util.concurrent.ConcurrentSkipListSet[String]()
```
tfenne: This used to be in dagr's `Configuration` singleton, so that it tracked the requested keys globally across all users of config regardless of where the traits were mixed in.  I don't think it's helpful to have it in the trait where it will only track the requests from each _instance_ that it's mixed into.  Maybe this can just go away here?

build.sbt
```diff
@@ -115,7 +115,8 @@ lazy val root = Project(id="commons", base=file("."))
   .settings(description := "Scala commons for Fulcrum Genomics.")
   .settings(
     libraryDependencies ++= Seq(
-      "org.scala-lang" %  "scala-reflect" %  scalaVersion.value,
+      "com.typesafe"   %  "config"        % "1.3.2",
```
tfenne, line 118: I guess this is fine, but I think we should be really careful about adding dependencies to `commons`.  It's a small project of utility classes, and even in this case the dependencies are now larger than the commons.jar we produce :/

tfenne (changes requested):
I think this PR needs some discussion as it introduces some significant changes vs. the version in dagr.  And some of the things that have been carried over made sense to me when it was an internal mechanism for dagr classes to access config, but don't make sense as a general config wrapper.

nh13:
@tfenne I took your suggested implementation, updated the documentation, and fixed a two things:
1. the `fetch` method should chain exceptions from its call to `asType`.
2. the `def requestedKey` was not tested and would produce a stack overflow.

Care to take one final look?

tfenne (approved):
Looks good.  Thanks for catching my mistakes!

The only thought I have, which is trivial, is that I think we originally used a ConcurrentSkipListSet so that we'd have the keys requested in the order they are requested, and of course scala doesn't have an insertion-order-maintaining set.  But then the `requstedKeys` method returns them in a sorted set (TreeSet).  So we may as well use a scala `mutable.Set()` instead of the java one.

PR #21, opened 2017-12-15
Add a method foldSum on collections to sum without creating intermediate

nh13:
collections.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/commons/CommonsDef.scala
+++ b/src/main/scala/com/fulcrumgenomics/commons/CommonsDef.scala
@@ -233,6 +233,24 @@ class CommonsDef {
   }
 
   /**
+    * Implicit that provides additional methods to sum the values of a sequence after applying a mapping to the items.
+    * This has the benefit of not creating any intermediate collections or iterators.   The mapping must map to a
+    * [[Numeric]] type.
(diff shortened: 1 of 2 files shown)


nh13:
@tfenne bump

src/main/scala/com/fulcrumgenomics/commons/CommonsDef.scala
```diff
@@ -232,6 +232,24 @@ class CommonsDef {
...
+    */
+  implicit class FoldSum[A](sequence: TraversableOnce[A]) {
+    /**
+      * Applies the given transform to the items in sequence, and then sums them.
+      * @param f the transform to map items from type [[A]] to type [[B]].
+      * @param x the numeric type of [[B]].
+      * @tparam B the type of the items after the transform has been applied; must be a [[Numeric]] type.
+      * @return
+      */
+    def foldSum[B](f: A => B)(implicit x: Numeric[B]): B = sequence.foldLeft(x.zero) { case (left, right) => x.plus(left, f(right)) }
```
tfenne: I like this, except how about calling it `sumBy`?  I think that's a little more in line with `maxBy` and `minBy` which exist in the standard API and do a mathematical operation over some transform of a sequence.

tfenne (approved):

PR #22, opened 2017-12-29
Adding some generic utility classes related to asynchronous processin…

nh13:
…g, reading, writing, and iterating over items.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,5 +1,4 @@
 import com.typesafe.sbt.SbtGit.GitCommand
-import sbt.Keys._
 import sbt._
 import sbtassembly.AssemblyKeys.assembly
 import sbtrelease.ReleasePlugin.autoImport.ReleaseTransformations._
--- a/src/main/scala/com/fulcrumgenomics/commons/async/Async.scala
+++ b/src/main/scala/com/fulcrumgenomics/commons/async/Async.scala
@@ -1,0 +1,385 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2017 Fulcrum Genomics LLC
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package com.fulcrumgenomics.commons.async
+
+import java.io.Closeable
+import java.util.concurrent.atomic.{AtomicBoolean, AtomicInteger, AtomicReference}
+import java.util.concurrent.{ArrayBlockingQueue, BlockingQueue, TimeUnit}
+
+import com.fulcrumgenomics.commons.io.Writer
+
+import scala.concurrent.forkjoin.ForkJoinPool
+
+/**
+  * Provides generic utility classes related to asynchronous processing, reading, writing, and iterating over items.
+  */
+object Async {
+
+  /** The default # of items to buffer in queues. */
+  private val DefaultBufferSize: Int = 2048
+
+
+  /** A trait with helper methods that can propagate exceptions from the underlying asynchronous thread.
+    *
+    * A reference to the underlying thread is given in `thread`. The thread will execute the run() method in a try block.
+    * If an exception occurs, throwable will be set if not already defined and uponException will be executed.  Finally,
+    * the uponFinally method will be executed.  The thread is set as a daemon but not started.
+    *
+    * The `checkAndRethrow` method should be used to check if the asynchronous thread has encountered an exception, and if
+    * so, rethrow the exception as an [[Error]] or [[RuntimeException]].
+    *
+    * The `tryAndModifyInterruptedException` method should be used on a block of code that blocks, and will throw a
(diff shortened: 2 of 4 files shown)


src/main/scala/com/fulcrumgenomics/commons/util/Async.scala
```diff
@@ -0,0 +1,286 @@
...
+    * @tparam T the type of object to process
+    */
+  class AsyncSink[T](val sink: T => Unit, bufferSize: Int = DefaultBufferSize, val source: Option[{ def close(): Unit }] = None)
+    extends Closeable {
+    require(bufferSize > 0, "bufferSize must be greater than zero")
+    private[util] val isClosed: AtomicBoolean = new AtomicBoolean(false)
+    private val queue: BlockingQueue[T] = new ArrayBlockingQueue[T](bufferSize)
+    private[util] val throwable: AtomicReference[Option[Throwable]] = new AtomicReference[Option[Throwable]](None)
+    private val sinkRunnable: Runnable = new AsyncSinkRunnable()
+    private[util] val sinkThread: Thread = {
```
nh13: does not need `private[util]`

src/main/scala/com/fulcrumgenomics/commons/util/Async.scala
```diff
@@ -0,0 +1,286 @@
...
+    *
+    * @param sink the method to invoke to process an object of type [[T]].
+    * @param bufferSize the number of elements to buffer before blocking when processing the elements
+    * @param source the optional source to close when this sink is closed.
+    * @tparam T the type of object to process
+    */
+  class AsyncSink[T](val sink: T => Unit, bufferSize: Int = DefaultBufferSize, val source: Option[{ def close(): Unit }] = None)
+    extends Closeable {
+    require(bufferSize > 0, "bufferSize must be greater than zero")
+    private[util] val isClosed: AtomicBoolean = new AtomicBoolean(false)
```
nh13: does not need private[util]

nh13:
@tfenne bump

nh13:
@tfenne bump

I would like to have this in commons so I can use it in fgbio.

src/main/scala/com/fulcrumgenomics/commons/util/Async.scala
```diff
@@ -0,0 +1,286 @@
...
+
+  /** The default # of items to buffer in queues. */
+  private val DefaultBufferSize: Int = 2048
+
+  private object AsyncIterator {
+    /** A [[Runnable]] who can be asked when it is all done.
+      *
+      * @param f the method to run
+      */
+    case class Runner(f: () => Unit) extends Runnable {
```
tfenne: I don't think this class is need and it obfuscates what's going on.  In can just have the `allDone` member inside `AsyncIterator` and declare runnable as:

```scala
val runnable = () => { outer.source.foreach(outer.queue.put); allDone.set(true) }
```

src/main/scala/com/fulcrumgenomics/commons/util/Async.scala
```diff
@@ -0,0 +1,286 @@
...
+    * @param source the source iterator to consume
+    * @param bufferSize the maximum number of elements to buffer before the blocking when consuming the source
+    * @tparam T the type of object to consume
+    */
+  class AsyncIterator[T](val source: Iterator[T], bufferSize: Int = DefaultBufferSize) extends Iterator[T] {
+    require(bufferSize > 0, "bufferSize must be greater than zero")
+    private val queue: BlockingQueue[T] = new ArrayBlockingQueue[T](bufferSize)
+    private val outer = this
+    private val runnable = AsyncIterator.Runner(() => outer.source.foreach(outer.queue.put))
+    private val thread = {
```
tfenne: You need to handle propogating exceptions from this thread back into the calling thread.  E.g. if this is consuming from an Iterator reading from a file and an IoException is thrown, the caller should see that exception the next time `next` of `hasNext` is called on `AsyncIterator`.

src/main/scala/com/fulcrumgenomics/commons/util/Async.scala
```diff
@@ -0,0 +1,286 @@
...
+      val t = new Thread(runnable, getClass.getSimpleName)
+      t.setDaemon(true)
+      t.start()
+      t
+    }
+
+    /** Returns true if there exists more elements, false otherwise */
+    def hasNext(): Boolean = {
+      if (!this.queue.isEmpty || this.source.nonEmpty) true
```
tfenne: I don't think you should ever touch `source` outside of `runnable`.  What if `nonEmpty` is side-effecting, e.g. it tries to queue up the next item?  At the very least you need to add synchronization so that here and in the runnable you can't be hitting source at the same time, but I'm not sure that's sufficient.

src/main/scala/com/fulcrumgenomics/commons/util/Async.scala
```diff
@@ -0,0 +1,286 @@
...
+    * in that group.
+    *
+    * @param writers the writers to which to write
+    * @param bufferSize the number of elements to buffer for each group of writers, before we block writing to writers
+    *                   in that group
+    * @param parallelism the maximum parallelism to use
+    * @param toIndex a method that returns the index into `writers` for each element to be written
+    * @tparam T the type of object to write
+    */
+  class AsyncGroupingWriter[T](writers: Traversable[Writer[T]],
```
tfenne: I don't really like this model.  It has a bunch of shortcomings, not least of which are:

1. The parallelism is silod.  E.g. if you have 96 writers, and one partition or group is getting 90% of the data, you're not going to see much benefit.
2. It requires you to be able to determine the writer from the thing being written which seems very task-specific

I think for something like this to live in commons we probably want it to be more general, and not have bottlenecks based on how data is distributed among the writers.

tfenne (changes requested):
A few major things:

1. I think the async iterator stuff has a bunch of problems.
2. Could you please move everything into an "async" package?  I think we should aim to not have a `util` package in `commons` over time as technically most of commons is utility code.
3. I have an aversion to the grouping writer.  Could that part live internal to DemuxFastqs for now?

nh13:
@tfenne I removed the `AsyncGroupingWriter` class as you suggested.  To support the propagation of exceptions in the underlying asynchronous thread in these classes, I pulled out some methods into `AsyncThread`.  This means that both `AsyncIterator` and `AsyncSink` use the same methods and runnable pattern to propagate exceptions from the underlying asynchronous thread.

version.sbt
```diff
@@ -4,4 +4,4 @@ gitHeadCommitSha in ThisBuild := Process("git rev-parse --short HEAD").lines.hea
 // *** IMPORTANT *** 
 // One of the two "version" lines below needs to be uncommented.
 // version in ThisBuild := "0.5.0" // the release version
- version in ThisBuild := s"0.5.0-${gitHeadCommitSha.value}-SNAPSHOT" // the snapshot version
+version in ThisBuild := s"0.5.0-${gitHeadCommitSha}-SNAPSHOT" // the snapshot version
```
nh13: @tfenne can you test this on your system?  I am not sure why I need this on mine.

nh13:
@tfenne I added a second attempt at a class that can asynchronously write across `N` writers using `M` threads (see `AsyncMultiWriter`).  For my specific use case, it is also important that the `close()` method occurs asynchronously across the writers, since in my case, I am using multiple `SamWriter`s, which each will sort and write to the final file only when `close()` is called.

nh13:
bump

nh13:
Closing in favor of #34.

PR #24, opened 2018-02-08
Fix ReflectionUtil's ability to create Chars.

tfenne:


src/main/scala/com/fulcrumgenomics/commons/reflect/ReflectionUtil.scala
```diff
@@ -376,6 +378,9 @@ object ReflectionUtil {
     else if (clazz == classOf[Path]) {
       PathUtil.pathTo(value)
     }
+    else if (clazz == classOf[java.lang.Character]) {
+      if (value.length == 0) 0.toChar else value.charAt(0)
```
tfenne: I'm not really sure what the right thing to do here is if the string is not length 1.  Suggestions?

nh13: My thought would be it _must_ be a string of length one.  I am not convinced about allowing an empty string, since there is no such thing as an empty character.

nh13 (changes requested):
I would just enforce the length of the string be one.

nh13 (approved):

PR #26, opened 2018-02-26
WIP: Example of converting to using mill for building.

tfenne:


nh13:
@tfenne I did some work on this, but I think the blocker is figuring out code-coverage.

nh13:
@tfenne I like Mill.  I have a working example of code coverage.  I think it belongs as its own seperate project (`mill-scoverage`?), so I'll create that.

tfenne:
Awesome @nh13.  Might it be worth opening an issue in `mill` to see if the author would accept a PR to add code coverage?  His philosophy on mill appears to be that things 90% of developers would need shouldn't have to be plugins, and I would have thought code coverage would fall into that, and I'm not sure there's any real alternative to `scoverage` in the scala ecosystem.

nh13:
@tfenne exactly. I have it working but I want to get it like Tests.

nh13:
Closing in favor of https://github.com/fulcrumgenomics/commons/pull/48

PR #27, opened 2018-03-04
Made Io.writeLines take TraversableOnce instead of Seq.

tfenne:


nh13 (approved):
It’s really hard to nitpick here. Any chance you could me something g to work with? :)

PR #28, opened 2018-03-06
Improve ReflectionUtil error messages.

nh13:
When no suitable one-arg string constructor was found, previously a
ClassNotFoundException was thrown with no helpful message.  Now we throw
a ReflectionException with the reason.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/commons/reflect/ReflectionUtil.scala
+++ b/src/main/scala/com/fulcrumgenomics/commons/reflect/ReflectionUtil.scala
@@ -387,7 +387,7 @@ object ReflectionUtil {
     }
     else if (clazz == classOf[Option[_]]) {
       // unitType shouldn't be an option, since it's supposed to the type *inside* any containers
-      throw new ReflectionException(s"Cannot construct an '${unitType.getSimpleName}' from a string.  Is this an Option[Option[...]]?")
+      throw new ReflectionException(s"Is this an Option[Option[...]]? Cannot construct an '${unitType.getSimpleName}'from a string: $value")
     }
     else if (clazz == classOf[java.lang.Object]) value
     else {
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/commons/reflect/ReflectionUtil.scala
```diff
@@ -387,7 +387,7 @@ object ReflectionUtil {
     }
     else if (clazz == classOf[Option[_]]) {
       // unitType shouldn't be an option, since it's supposed to the type *inside* any containers
-      throw new ReflectionException(s"Cannot construct an '${unitType.getSimpleName}' from a string.  Is this an Option[Option[...]]?")
+      throw new ReflectionException(s"Is this an Option[Option[...]]? Cannot construct an '${unitType.getSimpleName}'from a string: $value")
```
nh13, line 390: I added this in, since I want to know the string value too to help me figure out which one it was!

src/main/scala/com/fulcrumgenomics/commons/reflect/ReflectionUtil.scala
```diff
@@ -396,13 +396,15 @@ object ReflectionUtil {
         val ctor: Constructor[_] = clazz.getDeclaredConstructor(classOf[String])
         ctor.setAccessible(true)
         ctor.newInstance(value)
-      }.getOrElse {
+      }.orElse( Try {
```
tfenne: I'm a little confused here, and think this probably needs tightening up.  E.g. what if in the above block there _is_ a String constructor but the value is invalid?  E.g. `clazz` == Double and values == `"hello"`.

src/main/scala/com/fulcrumgenomics/commons/reflect/ReflectionUtil.scala
```diff
@@ -396,13 +396,15 @@ object ReflectionUtil {
...
         // Last chance: if everything else fails look to see if there is a companion object
         // with an apply(String) method that could work
         val typ             = mirror.classSymbol(clazz).toType
         val companionMirror = mirror.reflectModule(typ.typeSymbol.companion.asModule)
         val companion       = companionMirror.instance
         companion.getClass.getMethod("apply", classOf[String]).invoke(companion, value)
+      }).getOrElse {
+        throw new ReflectionException(s"No suitable single-argument constructor found to construct a '${unitType.getSimpleName}' from a string: $value")
```
tfenne: This isn't necessarily true - the constructor could have existed, but the call failed.

tfenne (changes requested):
Thanks for cleaning this up!  Mind closing the last little hole?

nh13:
@tfenne one last look?

tfenne (approved):
Thanks!

PR #29, opened 2018-04-04
Make Io.readLinesFromResource more friendly.

tfenne:
`getClass().getResourceAsStream()` and `getClass().getClassLoader().getResourceAsStream()` have different behaviours that can cause a lot of confusion.  Specifically:
  - the latter treats all paths as absolute but fails if you have a leading slash
  - the former treats paths as relative _unless_ they start with a slash

The result of this change is that a resource will be found if it is either:
a) A correct relative path
b) An absolute path without a slash
c) An absolute path with a slash

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/commons/io/Io.scala
+++ b/src/main/scala/com/fulcrumgenomics/commons/io/Io.scala
@@ -26,6 +26,7 @@ package com.fulcrumgenomics.commons.io
 import java.io._
 import java.nio.file.{Files, Path}
 
+import com.fulcrumgenomics.commons.CommonsDef._
 import scala.io.Source
 
 /**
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/commons/io/Io.scala
```diff
@@ -182,9 +182,14 @@ trait IoUtil {
 
   /** Finds a resource with a given name on the classpath and produces an iterator of lines of text from the resource. */
   def readLinesFromResource(name: String): Iterator[String] = {
-    val in = getClass.getResourceAsStream(name)
-    require(in != null, s"Resource does not exist at path: $name")
-    val stream = new BufferedInputStream(in, bufferSize)
-    Source.fromInputStream(stream).withClose(() => stream.close()).getLines
+    Seq(getClass.getResourceAsStream _, getClass.getClassLoader.getResourceAsStream _)
```
nh13, line 197: Insert a similar comment to the one you have in your PR description.

src/main/scala/com/fulcrumgenomics/commons/io/Io.scala
```diff
@@ -182,9 +182,14 @@ trait IoUtil {
...
+      .flatMap(m => Option(m(name)))
+      .headOption match {
+      case None     =>
+        throw new IllegalArgumentException(s"Resource does not exist at path: $name")
+      case Some(in) =>
+        val stream = new BufferedInputStream(in, bufferSize)
+        Source.fromInputStream(stream).withClose(() => stream.close()).getLines
```
nh13: Missing indent on 188-192.

nh13: While we're here, lets change `stream.close()` to `stream.safelyClose()` since it is an input stream?

src/test/scala/com/fulcrumgenomics/commons/io/IoTest.scala
```diff
@@ -164,12 +164,22 @@ class IoTest extends UnitSpec {
...
+
+  it should "work without a leading slash" in {
+    val actual = Io.readLinesFromResource("com/fulcrumgenomics/commons/io/to-lines-from-resource-test.txt").toList
+    actual shouldBe ResourceLines
+  }
 
-    actual shouldBe expected
+  it should "work with a relative path" in {
+    // This works because the resource is in the same package as this test class
+    val actual = Io.readLinesFromResource("to-lines-from-resource-test.txt").toList
```
nh13, line 182: Neat!

src/test/scala/com/fulcrumgenomics/commons/io/IoTest.scala
```diff
@@ -164,12 +164,22 @@ class IoTest extends UnitSpec {
     roundtripped shouldBe lines
   }
 
+  val ResourceLines = Io.readLines(Paths.get("src/test/resources/com/fulcrumgenomics/commons/io/to-lines-from-resource-test.txt")).toList
```
nh13: private

nh13 (dismissed):
Nitpicky comments but nothing I need to re-review.

PR #31, opened 2018-04-05
Upgrading the build to sbt 1.1.2.

tfenne:
Another housekeeping PR @nh13


build.sbt
```diff
@@ -121,10 +120,3 @@ lazy val root = Project(id="commons", base=file("."))
       "org.scalatest"  %% "scalatest"     % "3.0.1"  % "test->*" excludeAll ExclusionRule(organization="org.junit", name="junit")
     )
   )
-
-lazy val shadowJar = TaskKey[Unit]("shadowJar", "Creates an assembly JAR without the trailing git hash and SNAPSHOT in the name")
```
tfenne, line 125: I feel like this is vestigal, especially in commons where a shadow/assembly jar is largely useless.  If I'm wrong let me know and I'll restore it.

project/plugins.sbt
```diff
@@ -1,9 +1,9 @@
 resolvers += Resolver.url("fix-sbt-plugin-releases", url("http://dl.bintray.com/sbt/sbt-plugin-releases"))(Resolver.ivyStylePatterns)
 
-addSbtPlugin("com.typesafe.sbt"  % "sbt-git"       % "0.9.2")
-addSbtPlugin("com.github.gseitz" % "sbt-release"   % "1.0.3")
-addSbtPlugin("com.eed3si9n"      % "sbt-assembly"  % "0.14.3")
-addSbtPlugin("org.scoverage"     % "sbt-scoverage" % "1.5.0")
-addSbtPlugin("org.xerial.sbt"    % "sbt-sonatype"  % "1.1")
-addSbtPlugin("com.jsuereth"      % "sbt-pgp"       % "1.0.1")
-addSbtPlugin("com.eed3si9n"      % "sbt-unidoc"    % "0.3.3")
```
nh13, line 9: Do we use unidoc?  If not, then this is fine.

nh13 (approved):

project/plugins.sbt
tfenne: Nah.  I think it's a holdover from when these were all sub-projects in one big `dagr` project.

PR #33, opened 2018-07-30
Added Io.readBytesFromResources() to read binary data.

tfenne:


src/main/scala/com/fulcrumgenomics/commons/io/Io.scala
```diff
@@ -191,17 +200,27 @@ trait IoUtil {
...
-      .flatMap(m => Option(m(name)))
-      .headOption match {
-        case None     =>
-          throw new IllegalArgumentException(s"Resource does not exist at path: $name")
-        case Some(in) =>
-          val stream = new BufferedInputStream(in, bufferSize)
-          Source.fromInputStream(stream).withClose(() => stream.safelyClose()).getLines
-    }
+    val stream = new BufferedInputStream(streamFromResource(name), bufferSize)
+    Source.fromInputStream(stream).withClose(() => stream.safelyClose()).getLines
```
nh13, line 205: Thanks for this too!

src/main/scala/com/fulcrumgenomics/commons/io/Io.scala
```diff
@@ -27,6 +27,8 @@ import java.io._
 import java.nio.file.{Files, Path}
 
 import com.fulcrumgenomics.commons.CommonsDef._
+
+import scala.collection.mutable.ArrayBuffer
```
nh13: unused import

nh13 (approved):
One unneeded import, otherwise LGTM.

src/main/scala/com/fulcrumgenomics/commons/io/Io.scala
tfenne: Doh, thanks!

PR #34, opened 2018-08-15
A number of updates

nh13:
This has three major components:
1. Async utility classes.  This PR supersedes https://github.com/fulcrumgenomics/commons/pull/22
2. A `StringUtil.split` method for faster String splitting
3. Adding a `+=` method to `SimpleCounter`.

Please see the review comments in #22 for the Async commit.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/commons/CommonsDef.scala
+++ b/src/main/scala/com/fulcrumgenomics/commons/CommonsDef.scala
@@ -25,11 +25,13 @@
 package com.fulcrumgenomics.commons
 
 import java.io.Closeable
+import java.util.Comparator
 
 import com.fulcrumgenomics.commons.collection.BetterBufferedIterator
 import com.fulcrumgenomics.commons.util.Logger
 
-import scala.collection.Parallelizable
+import scala.collection.mutable.ListBuffer
+import scala.collection.{LinearSeq, LinearSeqOptimized, Parallelizable, SeqLike}
 import scala.collection.parallel.{ForkJoinTaskSupport, ParIterableLike, TaskSupport}
 import scala.concurrent.forkjoin.ForkJoinPool
 import scala.language.implicitConversions
@@ -134,6 +136,99 @@ class CommonsDef {
       } catch {
         case e: Exception => logger.foreach(_.exception(e))
       }
+    }
+  }
+
+  implicit class MaxNBy[A](val things: TraversableOnce[A]) {
+    /** Finds the first `n` elements with the largest values.
+      *
+      * @param n the number of elements return.
+      * @param distinct true to return the first `n` distinct elements, otherwise false to allow duplicate elements.
+      * @param cmp the implicit comparison method for type [[A]].
+      * @return  the first `n` elements of this traversable or iterator with the largest values.
+      */
+    def maxN(n: Int, distinct: Boolean = false)(implicit cmp: Ordering[A]): Seq[A] = this.maxNBy[A](n, thing => thing, distinct)
+
+    /** Finds the first `n` elements with the largest values measured by function f.
+      *
+      * @param n the number of elements return.
+      * @param f the measuring function.
+      * @param distinct true to return the first `n` distinct elements, otherwise false to allow duplicate elements.
+      * @param cmp the implicit comparison method for type [[B]].
+      * @tparam B the result type of the function f.
+      * @return  the first `n` elements of this traversable or iterator with the largest values measured by function f.
+      */
+    def maxNBy[B](n: Int, f: A => B, distinct: Boolean = false)(implicit cmp: Ordering[B]): Seq[A] = if (things.isEmpty || n == 0) Seq.empty[A] else {
+      // Developer Note: a future improvement would be to determine when to use maxNBySmall vs. maxNByLarge.  Also,
+      // if we want to get n items from a sequence that is close to length m, we could find the m-n items to remove, as
+      // to save memory.
+      things match {
+        case _things: Seq[A] => if (_things.length <= n && n < 1024) maxNBySmall(_things, n, f, distinct) else  maxNByLarge(n, f, distinct)
+        case _               => maxNByLarge(n, f, distinct)
+      }
+    }
+
+    /** Companion to [[ThingAndValue]] that provides various apply methods. */
+    private object ThingAndValue {
+      def apply[B](thing: A, value: B)(implicit cmp: Ordering[B]): ThingAndValue[B] = new UnindexedThingAndValue[B](thing, value)
+      def apply[B](thing: A, value: B, index: Int)(implicit cmp: Ordering[B]): ThingAndValue[B] = new IndexedThingAndValue[B](thing, value, index)
+    }
+
+    /** The base trait for storing things and values, where the "thing"s can ordered by the "value"s. */
+    private sealed trait ThingAndValue[B] extends Comparable[ThingAndValue[B]] {
+      implicit def cmp: Ordering[B]
+      def thing: A
+      def value: B
+      override def compareTo(that: ThingAndValue[B]): Int = cmp.compare(that.value, this.value) // sort by the maximum
+
+    }
+
+    /** A little class to store the "things" and their "values". */
+    private case class UnindexedThingAndValue[B](thing: A, value: B)(implicit val cmp: Ordering[B]) extends ThingAndValue[B]
+
+    /** A little class to store the "things" and their "values", as well as their index in the original collection.  We
+      * use the index to distinguish between duplicates. */
+    private case class IndexedThingAndValue[B](thing: A, value: B, index: Int)(implicit val cmp: Ordering[B]) extends ThingAndValue[B] {
+      override def compareTo(that: ThingAndValue[B]): Int = that match {
+        case _that: IndexedThingAndValue[B] =>
+          // sort by maximum value first, then by smaller index
+          cmp.compare(_that.value, this.value) match {
+            case 0 => this.index.compareTo(_that.index)
+            case r => r
+          }
+        case _ => super.compareTo(that)
+      }
+    }
+
+    /** Implementation of [[maxNBy()]] for when the # of elements is "small". */
+    @inline
+    private def maxNBySmall[B](things: Seq[A], n: Int, f: A => B, distinct: Boolean = false)(implicit cmp: Ordering[B]): Seq[A] = {
+      if (distinct) things.map(t => ThingAndValue(t, f(t))).sorted.map(_.thing).distinct
+      else things.zipWithIndex.map { case (t, i) => ThingAndValue(t, f(t), i) }.sorted.map(_.thing)
+    }
+
+    /** Implementation of [[maxNBy()]] for when the # of elements is "large". */
+    @inline
+    private def maxNByLarge[B](n: Int, f: A => B, distinct: Boolean = false)(implicit cmp: Ordering[B]): Seq[A] = {
+      val orderedThings = new java.util.TreeSet[ThingAndValue[B]]()
+      val toThingAndValue: (A, Int) => ThingAndValue[B] = {
+        if (distinct) (t: A, _: Int) => ThingAndValue(t, f(t)) else (t: A, i: Int) => ThingAndValue(t, f(t), i)
+      }
+      this.things.toIterator.zipWithIndex.foreach { case (thing, index) =>
+        val thingAndValue = toThingAndValue(thing, index)
+        // If we haven't reached N yet, add it.  If the last element is "smaller" than the curret one, add it.  Otherwise, don't!
+        if (orderedThings.isEmpty || orderedThings.size() < n) orderedThings.add(thingAndValue)
+        else {
+          // The last thing should exist, so compare it to the current
+          val lastThingAndValue = orderedThings.last()
+          if (lastThingAndValue.compareTo(thingAndValue) > 0) {
+            // Only remove the last thing if the new thing was added
+            if (orderedThings.add(thingAndValue)) orderedThings.pollLast()
+          }
+        }
+        require(orderedThings.size <= n, s"$n ${orderedThings.toList}")
+      }
+      orderedThings.map(_.thing).toList
     }
   }
 
--- a/src/main/scala/com/fulcrumgenomics/commons/async/AsyncIterator.scala
+++ b/src/main/scala/com/fulcrumgenomics/commons/async/AsyncIterator.scala
@@ -1,0 +1,89 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2018 Fulcrum Genomics
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ */
+
+package com.fulcrumgenomics.commons.async
+
+import java.util.concurrent.{ArrayBlockingQueue, BlockingQueue, LinkedBlockingDeque, TimeUnit}
+
+object AsyncIterator {
+  /** Builds an [[AsyncIterator]] and starts it.
+    *
+    * @param source the source iterator to consume
+    * @param bufferSize the maximum number of elements to buffer before the blocking when consuming the source, or None
+    *                   if unbounded.
+    * @tparam T the type of object to consume
+    */
+  def apply[T](source: Iterator[T], bufferSize: Option[Int] = None): AsyncIterator[T] = {
+    new AsyncIterator[T](source, bufferSize).start()
+  }
+}
+
+/** An asynchronous wrapper for an [[Iterator]].  A separate thread will be created to consume the source iterator.
+  * Will buffer up to [[bufferSize]] elements before the blocking when consuming the source.
+  *
+  * @param source the source iterator to consume
+  * @param bufferSize the maximum number of elements to buffer before the blocking when consuming the source, or None
+  *                   if unbounded.
+  * @tparam T the type of object to consume
+  */
+class AsyncIterator[T](private val source: Iterator[T], bufferSize: Option[Int] = None) extends Iterator[T] with AsyncRunnable {
+  bufferSize.foreach(size => require(size > 0, s"bufferSize must be greater than zero when given, found $size"))
+
+  private var buffer: Option[T] = None
+
+  private val queue: BlockingQueue[T] = bufferSize match {
+    case Some(size) => new ArrayBlockingQueue[T](size)
+    case None       => new LinkedBlockingDeque[T]()
+  }
+
+  protected def execute(): Unit = this.source.foreach(this.queue.put)
+
+  /** Returns true if there exists more elements, false otherwise */
+  def hasNext(): Boolean = {
+    checkAndRaise()
+
+    // Get the next item, or wait until the underlying thread is done and there are no more items in the queue
+    while (buffer.isEmpty && !(this.done && this.queue.isEmpty)) {
+      checkAndRaise() // check if hte underlying thread raised an exception
+      tryAndModifyInterruptedException("Interrupted waiting on taking from the queue.") {
+        buffer = Option(this.queue.poll(50, TimeUnit.MILLISECONDS))
+      }
+    }
+
+    // Did we get an item from the buffer?
+    buffer.nonEmpty
+  }
+
+  /** Gets the next item. */
+  def next(): T = {
+    checkAndRaise()
+    this.buffer match {
+      case None => throw new NoSuchElementException("Calling next() when hasNext() is false.")
+      case Some(item) =>
+        this.buffer = None
+        item
+    }
+  }
+}
--- a/src/main/scala/com/fulcrumgenomics/commons/async/AsyncRunnable.scala
+++ b/src/main/scala/com/fulcrumgenomics/commons/async/AsyncRunnable.scala
@@ -1,0 +1,163 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2018 Fulcrum Genomics
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ */
+
+package com.fulcrumgenomics.commons.async
+
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.atomic.{AtomicInteger, AtomicReference}
+
+import scala.concurrent.{Awaitable, ExecutionContext, Future}
+
+object AsyncRunnable {
+  /** A counter used for thread naming in [[AsyncRunnable]] and sub-classes */
+  private val threadsCreated = new AtomicInteger(0)
+
+  /** Generates a name for thread that will run the given runnable. */
+  private[AsyncRunnable] def nextName(asyncRunnable: AsyncRunnable): String = {
+    asyncRunnable.getClass.getSimpleName + this.threadsCreated.incrementAndGet
+  }
+}
+
+/** A trait that can be mixed in to help manage asynchronous execution while being [[Runnable]].
+  *
+  * The `execute()` method should contain the work to be performed asynchronously.
+  *
+  * The `thread()` method can be used to return a [[Thread]] that wraps this [[Runnable]].
+  *
+  * The `start()` method can be used to create and start daemon [[Thread]] that wraps this [[Runnable]].
+  *
+  * The `checkAndRaise()` method should be used by implementing classes to check if the
+  * `execute()`  method has encountered an exception, and if so, it will rethrow the exception as an
+  * [[Error]] or [[RuntimeException]].
+  *
+  * The `tryAndModifyInterruptedException()` method should wrap blocking code in implementing classes
+  * that are not part of the `execute()` method.  This method will throw a [[RuntimeException]] if interrupted while
+  * waiting.
+  */
+trait AsyncRunnable extends Runnable {
+
+  /** A reference to an exception thrown during an asynchronous operation. */
+  private[AsyncRunnable] val _throwable: AtomicReference[Throwable] = new AtomicReference(null)
+
+  /** This signals that the run method has started. */
+  private val startedLatch: CountDownLatch = new CountDownLatch(1)
+
+  /** This signals that the run method has completed, even if an exception occurred. This is needed so the close method
+    * can be sure that the run method is no longer executing. */
+  private val doneLatch: CountDownLatch = new CountDownLatch(1)
+
+  final def run(): Unit = {
+    startedLatch.countDown()
+    try {
+      this.execute()
+    } catch {
+      case thr: Throwable =>
+        this._throwable.compareAndSet(null, thr)
+        this.uponException()
+    } finally {
+      this.uponFinally()
+      this.doneLatch.countDown()
+    }
+  }
+
+  /** Returns true if the [[run()]] method has started, false otherwise. */
+  def started: Boolean = this.startedLatch.getCount == 0
+
+  /** Returns true if the [[run()]] method has completed, false otherwise. */
+  def done: Boolean = this.doneLatch.getCount == 0
+
+  /** Returns a throwable if an exception occurred in the [[run()]] method, None otherwise. */
+  def throwable: Option[Throwable] = Option(this._throwable.get())
+
+  /** Waits for the [[run()]] method to start. */
+  def awaitStart(): Unit = this.startedLatch.await()
+
+  /** Returns an [[Awaitable]] that completes when the [[run()]] method has started.  Returns the throwable if an
+    * exception was encountered, [[None]] otherwise. */
+  def uponStart()(implicit ec: ExecutionContext) : Awaitable[Unit] = Future { this.awaitStart() }
+
+  /** Waits for the [[run()]] method to complete. */
+  def awaitDone(): Unit = this.doneLatch.await()
+
+  /** Returns an [[Awaitable]] that completes when the [[run()]] method has completed.  Returns the throwable if an
+    * exception was encountered, [[None]] otherwise. */
+  def uponDone()(implicit ec: ExecutionContext) : Awaitable[Option[Throwable]] = Future { this.awaitDone(); Option(this._throwable.get()) }
+
+  /** The name of this runnable.  This is used as the name of the thread in [[thread()]] as well.  The name is created
+    * based on the class name and the number of [[AsyncRunnable]]s already created.
+    * */
+  val name: String = AsyncRunnable.nextName(this)
+
+  /** Creates a new thread wrapping this runnable; the thread is not started. */
+  final def thread(): Thread = {
+    require(!this.started, "Already started.")
+    new Thread(this, this.name)
+  }
+
+  /** Starts this [[Runnable]] in a daemon thread.
+    *
+    * @param name optionally the name of the thread, otherwise a name is created based on the class name and the number
+    *             of threads already created.
+    */
+  final def start(name: Option[String] = None, daemon: Boolean = true): this.type = {
+    require(!this.started, "Already started.")
+    val t = thread()
+    t.setDaemon(daemon)
+    t.start()
+    this.startedLatch.countDown()
+    this
+  }
+
+  /** The method that does the asynchronous work. */
+  protected def execute(): Unit
+
+  /** The method to execute if an exception occurs in the asynchronous thread.  This should not block. */
+  protected def uponException(): Unit = Unit
+
+  /** The method to execute upon successfully execution of the run method or an exception occurs.  This should not block. */
+  protected def uponFinally(): Unit = Unit
+
+  /** Checks to see if an exception has been raised by an asynchronous thread and if so rethrows it.  Use this method
+    * before code that assumes the threads have not encountered an exception.
+    */
+  protected final def checkAndRaise(): Unit = {
+    _throwable.getAndSet(null) match {
+      case null           => Unit
+      case thr: Throwable => throw thr
+    }
+  }
+
+  /** Executes the given block of code.  If an [[InterruptedException]] is thrown, throws a [[RuntimeException]]
+    * with the given message.  Use this method for blocking code that when interrupted should not be recoverable.
+    *
+    * @param message the message to use if an [[InterruptedException]] is thrown by the block of code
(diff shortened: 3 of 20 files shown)


src/test/scala/com/fulcrumgenomics/commons/util/StringUtilTest.scala
```diff
@@ -110,7 +110,18 @@ class StringUtilTest extends UnitSpec {
     StringUtil.addSpacesToCamelCase("b") should be ("b")
   }
 
-  it should "add a space between adjacen upper case characters" in {
```
nh13, line 113: this was my best contribution in this PR

src/main/scala/com/fulcrumgenomics/commons/CommonsDef.scala
```diff
@@ -137,6 +138,65 @@ class CommonsDef {
...
+  implicit class MaxNBy[A](things: TraversableOnce[A]) {
+    def maxN[B](n: Int, distinct: Boolean = false)(implicit cmp: Ordering[A]): Seq[A] = this.maxNBy[A](n, thing => thing, distinct)
+
+    def maxNBy[B](n: Int, f: A => B, distinct: Boolean = false)(implicit cmp: Ordering[B]): Seq[A] = {
+      require(0 < n, "n must be greater than zero")
+      // Developer Note: a future improvement would be to determine when to use maxNBySmall vs. maxNByLarge.  Also,
+      // if we want to get n items from a sequence that is close to length m, we could find the m-n items to remove, as
+      // to save memory.
+      things match {
+        case _things: Seq[A] if _things.lengthCompare(n) <= 0 && n < 1024 => maxNBySmall(_things, n, f, distinct)
```
tfenne: I had to look to see what `lengthCompare` does.  FWIW it creates an iterator and iterates through the seq until it either hits n or the end.  For `List` this is more efficient, but for any indexed structures with a known length, it's much less efficient.  I think this would be much clearer just as:

```scala
if (things.length <= n && n < 1024) maxNBySmall(...) else maxNByLarge(...)
```

src/main/scala/com/fulcrumgenomics/commons/CommonsDef.scala
```diff
@@ -137,6 +138,65 @@ class CommonsDef {
     }
   }
 
+  implicit class MaxNBy[A](things: TraversableOnce[A]) {
```
tfenne: Probably should put `val` in front of `things`.

src/main/scala/com/fulcrumgenomics/commons/CommonsDef.scala
```diff
@@ -137,6 +138,65 @@ class CommonsDef {
...
+    def maxN[B](n: Int, distinct: Boolean = false)(implicit cmp: Ordering[A]): Seq[A] = this.maxNBy[A](n, thing => thing, distinct)
```
tfenne: The two `maxN` and `maxNBy` could use scaladoc please.

src/main/scala/com/fulcrumgenomics/commons/CommonsDef.scala
```diff
@@ -137,6 +138,65 @@ class CommonsDef {
...
+      // to save memory.
+      things match {
+        case _things: Seq[A] if _things.lengthCompare(n) <= 0 && n < 1024 => maxNBySmall(_things, n, f, distinct)
+        case _                                                            => maxNByLarge(n, f, distinct)
+      }
+    }
+
+    @inline
+    private def maxNBySmall[B](things: Seq[A], n: Int, f: A => B, distinct: Boolean = false)(implicit cmp: Ordering[B]): Seq[A] = {
+      if (distinct) things.sortWith((a, b) => cmp.gteq(f(a), f(b))).distinct
```
tfenne: To your point, wouldn't this be more efficient as:

```scala
things.map(t => ThingAndValue(t, f(t)).sorted.map(_._1).distinct
```
?

src/main/scala/com/fulcrumgenomics/commons/CommonsDef.scala
```diff
@@ -137,6 +138,65 @@ class CommonsDef {
...
+          }
+          retval < 0
+        }.map(_._1)
+      }
+    }
+
+    @inline
+    private def maxNByLarge[B](n: Int, f: A => B, distinct: Boolean = false)(implicit cmp: Ordering[B]): Seq[A] = {
+      // A little class to store the "things" and their "values".  We use the index to distinguish between duplicates.
+      case class ThingAndValue(index: Int, thing: A, value: B) extends Comparable[ThingAndValue] {
```
tfenne: I think probably this should just be declared inside `class MaxNBy` not in the function.

src/main/scala/com/fulcrumgenomics/commons/CommonsDef.scala
```diff
@@ -137,6 +138,65 @@ class CommonsDef {
...
+        case _                                                            => maxNByLarge(n, f, distinct)
+      }
+    }
+
+    @inline
+    private def maxNBySmall[B](things: Seq[A], n: Int, f: A => B, distinct: Boolean = false)(implicit cmp: Ordering[B]): Seq[A] = {
+      if (distinct) things.sortWith((a, b) => cmp.gteq(f(a), f(b))).distinct
+      else {
+        things.zipWithIndex.sortWith { case ((a, ai), (b, bi)) =>
+          val retval = cmp.compare(f(b), f(a)) match {
```
tfenne: Is this necessary?  Is `sortWith` not stable?

src/main/scala/com/fulcrumgenomics/commons/CommonsDef.scala
```diff
@@ -137,6 +138,65 @@ class CommonsDef {
...
+      this.things.toIterator.zipWithIndex.foreach { case (thing, index) =>
+        val thingAndValue = ThingAndValue(if (distinct) 0 else index, thing, f(thing))
+        // If we haven't reached N yet, add it.  If the last element is "smaller" than the curret one, add it.  Otherwise, don't!
+        if (orderedThings.isEmpty || orderedThings.size() < n) orderedThings.add(thingAndValue)
+        else {
+          // The last thing should exist, so compare it to the currnt
+          val lastThingAndValue = orderedThings.last()
+          if (lastThingAndValue.compareTo(thingAndValue) > 0) {
+            // Only remove the last thing if the new thing was added
+            if (orderedThings.add(thingAndValue)) orderedThings.remove(lastThingAndValue)
```
tfenne: I think probably `orderedThings.pollLast()` will be a more efficient way of removing the last item.

src/main/scala/com/fulcrumgenomics/commons/async/Async.scala
```diff
@@ -0,0 +1,397 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2017 Fulcrum Genomics LLC
```
tfenne: 2018?

src/main/scala/com/fulcrumgenomics/commons/async/Async.scala
```diff
@@ -0,0 +1,397 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package com.fulcrumgenomics.commons.async
```
tfenne: I tend to think that if everything is packaged inside an `Async` object, that acts like the packages and it's not helpful to have a deeper package structure.

src/main/scala/com/fulcrumgenomics/commons/async/Async.scala
```diff
@@ -0,0 +1,397 @@
...
+    /** The method to execute if an exception occurs in the asynchronous thread. */
+    protected def uponException(): Unit =  Unit
+
+    /** The method to execute upon successfully execution of the run method or an exception occurs. */
+    protected def uponFinally(): Unit = Unit
+
+    /** The asynchronous daemon thread to execute.   */
+    protected def buildThread(name: String): Thread = {
+      // NB: implicit conversion to Runnable not supported in scala 2.11
+      val runnable = new Runnable {
```
tfenne: I think this would be easier to follow if `AsyncThread`:
1. extended `Runnable`
2. defined a `def final run(): Unit` method that has the try/catch/finally from here
3. defined an abstract `def execute(): Unit` (or similar) that subclasses implement

Alternatively you could also just make `AsyncThread` a class that takes in the name and a `f: () => Unit`, and constructs the thread to invoke `f`.

src/main/scala/com/fulcrumgenomics/commons/async/Async.scala
```diff
@@ -0,0 +1,397 @@
...
+    * If an exception occurs, throwable will be set if not already defined and uponException will be executed.  Finally,
+    * the uponFinally method will be executed.  The thread is set as a daemon but not started.
+    *
+    * The `checkAndRethrow` method should be used to check if the asynchronous thread has encountered an exception, and if
+    * so, rethrow the exception as an [[Error]] or [[RuntimeException]].
+    *
+    * The `tryAndModifyInterruptedException` method should be used on a block of code that blocks, and will throw a
+    * [[RuntimeException]] if the interrupted.
+    * */
+  private[async] trait AsyncThread {
```
tfenne: Maybe `SafeRunnable`? since it's not actually a `Thread`?

src/main/scala/com/fulcrumgenomics/commons/async/Async.scala
```diff
@@ -0,0 +1,397 @@
...
+          } catch {
+            case thr: Throwable =>
+              outer.throwable.compareAndSet(None, Some(thr))
+              outer.uponException()
+          } finally {
+            outer.uponFinally()
+          }
+        }
+      }
+      val t = new Thread(runnable, name)
```
tfenne: I think it might be nicer if it just retained access to the thread and had pass-through methods for things like `join()`, `isDone()` etc.  That way how it gets run asynchronously is hidden and could be changed later to use a `ForkJoinPool` or other mechanism.

src/main/scala/com/fulcrumgenomics/commons/async/Async.scala
```diff
@@ -0,0 +1,397 @@
...
+    protected def tryAndModifyInterruptedException[T](message: String)(f: => T): T = {
+      try { f } catch {
+        case ie: InterruptedException => throw new RuntimeException(message, ie)
+      }
+    }
+
+    /** Checks to see if an exception has been raised in the asynchronous thread and if so rethrows it as an [[Error]]
+      * or [[RuntimeException]] as appropriate.
+      */
+    protected final def checkAndRethrow(): Unit = {
```
tfenne: Shouldn't this be public?

tfenne: Also, I think the return type should be `Nothing` in this case.

src/main/scala/com/fulcrumgenomics/commons/async/Async.scala
```diff
@@ -0,0 +1,397 @@
...
+
+  /** An asynchronous wrapper for an [[Iterator]].  A separate thread will be created to consume the source iterator.
+    * Will buffer up to [[bufferSize]] elements before the blocking when consuming the source.
+    *
+    * @param source the source iterator to consume
+    * @param bufferSize the maximum number of elements to buffer before the blocking when consuming the source, or None
+    *                   if unbounded.
+    * @tparam T the type of object to consume
+    */
+  class AsyncIterator[T](val source: Iterator[T], bufferSize: Option[Int] = None) extends Iterator[T] with AsyncThread {
```
tfenne: `private val source`?  Also, I think I'd lean towards making `bufferSize` required.  Users can always set it to `Int.MaxValue` or some other stupidly large number if needed.

src/main/scala/com/fulcrumgenomics/commons/async/Async.scala
```diff
@@ -0,0 +1,397 @@
...
+    bufferSize.foreach(size => require(size > 0, s"bufferSize must be greater than zero when given, found $size"))
+
+    private val queue: BlockingQueue[T] = bufferSize match {
+      case Some(size) => new ArrayBlockingQueue[T](size)
+      case None       => new LinkedBlockingDeque[T]()
+    }
+    private val thread = buildThread(getClass.getSimpleName)
+    private val outer  = this
```
tfenne: This isn't needed; all `outer`s can be replaced with `this`.

src/main/scala/com/fulcrumgenomics/commons/async/Async.scala
```diff
@@ -0,0 +1,397 @@
...
+
+    // Start the thread after creating the queue
+    this.thread.start()
+
+    protected def run(): Unit = outer.source.foreach(outer.queue.put)
+
+    override protected def uponFinally(): Unit = outer.isDone.set(true)
```
tfenne: I think `isDone.set(true)` should be done in the parent trait, in the `finally` block, and not reliant upon sublcasses to set it.  Also, I think it should probably be `private` with a simple `def isDone: Boolean` accessor in case we decide to implement differently (e.g. `!thread.isAlive`)

src/main/scala/com/fulcrumgenomics/commons/async/Async.scala
```diff
@@ -0,0 +1,397 @@
...
+
+    /** Returns true if there exists more elements, false otherwise */
+    def hasNext(): Boolean = {
+      checkAndRethrow()
+
+      // If the queue is empty but the underlying thread is not done, then there may or may not be new
+      // items that will be put in the qeue.
```
tfenne: q[u]eue

src/main/scala/com/fulcrumgenomics/commons/async/Async.scala
```diff
@@ -0,0 +1,397 @@
...
+      while (this.queue.isEmpty && !this.isDone.get()) {
+        Thread.sleep(50)
```
tfenne: The `Thread.sleep()` with an arbitrary number here bothers me, but I'm struggling to suggest a better solution that's not much more complicated.  

One option would be to implement as:
`!this.queue.isEmpty || ! this.iterator.syncrhonized(this.iterator.isEmpty)`
but that would require synchronizing access to the iterator in the other thread.

Another option would be to have this also be a buffered iterator with a `_next: Option[T]` and have this be:

```scala
def hasNext(): Boolean = {
    while(next.isEmpty && !isDone) _next = Option(queue.poll(50, TimeUnit.Milliseconds))
  _next.isDefined
}
```

src/main/scala/com/fulcrumgenomics/commons/async/Async.scala
```diff
@@ -0,0 +1,397 @@
...
+        checkAndRethrow() // check if there was any exception in the underlying thread
+      }
+
+      // At this point, either there are items in the queue, or the underlying thread has consumed the source
+      if (!this.queue.isEmpty) true
+      else { // the queue is empty, and the source has been consumed
+        require(this.isDone.get(), "Bug: iterator was not done when the thread was joined.")
+        tryAndModifyInterruptedException("Interrupted waiting on iterator thread to join.") { this.thread.join() }
```
tfenne: Why do you need to join if `isDone` is already set?

src/main/scala/com/fulcrumgenomics/commons/async/Async.scala
```diff
@@ -0,0 +1,397 @@
...
+        !this.queue.isEmpty
+      }
+    }
+
+    /** Gets the next item. */
+    def next(): T = {
+      checkAndRethrow()
+      if (this.hasNext()) {
+        val item = tryAndModifyInterruptedException("Interrupted waiting on taking from the queue.") { this.queue.take() }
+        checkAndRethrow()
```
tfenne: I'm not sure you need to call this multiple times per method.  I think once at the start will do.

src/main/scala/com/fulcrumgenomics/commons/async/Async.scala
```diff
@@ -0,0 +1,397 @@
...
+    */
+  class AsyncSink[T](val sink: T => Unit, bufferSize: Option[Int] = None, val source: Option[{ def close(): Unit }] = None)
+    extends Closeable with AsyncThread {
+    bufferSize.foreach(size => require(size > 0, s"bufferSize must be greater than zero when given, found $size"))
+    private val queue: BlockingQueue[T] = bufferSize match {
+      case Some(size) => new ArrayBlockingQueue[T](size)
+      case None       => new LinkedBlockingDeque[T]()
+    }
+    private val thread = buildThread(this.getClass.getSimpleName + AsyncSink.threadsCreated.incrementAndGet)
+    private val outer = this
```
tfenne: Unneeded.

src/main/scala/com/fulcrumgenomics/commons/async/Async.scala
```diff
@@ -0,0 +1,397 @@
...
+      */
+    def close(): Unit = {
+      checkAndRethrow()
+      if (!this.isDone.getAndSet(true)) {
+        tryAndModifyInterruptedException("Interrupted waiting on sink thread.") { this.thread.join() }
+        // The queue should be empty but if it's not, we'll drain it here to protect against any lost data.
+        // There's no need to timeout on poll because take is called only when queue is not empty and
+        // at this point the thread is definitely dead and no-one is removing items from the queue.
+        // The item pulled will never be null (same reasoning).
+        while (!this.queue.isEmpty) {
```
tfenne: These suspenders don't go with this belt!  Why not just `require(this.queue.isEmpty)` here?

src/main/scala/com/fulcrumgenomics/commons/async/Async.scala
```diff
@@ -0,0 +1,397 @@
...
+    *                   if unbounded
+    * @param minimumWaitMillis the minimum time to sleep after unsuccessfully retrieving a non-empty queue of items
+    * @param maximumWaitMillis the maximum time to sleep after unsuccessfully retrieving a non-empty queue of items
+    * @param incrementWaitMillis the amount of time to increase the sleep time after unsuccessfully retrieving a
+    *                            non-empty queue of items
+    * @param decreaseWaitFactor the factor to scale the sleep time (0 < decreasePollFactor < 1) after successfully
+    *                           retrieving a non-empty queue of items
+    * @tparam T the type of items to write.
+    */
+  class AsyncMultiWriter[T](writers: Traversable[Writer[T]],
```
tfenne: I'm ignoring this class for now because I don't really understand the use case.

src/main/scala/com/fulcrumgenomics/commons/util/SimpleCounter.scala
```diff
@@ -63,4 +63,10 @@ class SimpleCounter[T]extends Iterable[(T, Long)] {
 
   /** Gets the number of counts stored in this counter (sum over all count). */
   def total: Long = _total
+
+  /** Adds the items and counts in `other` to the counts here. */
+  def +=(other: SimpleCounter[T]): this.type = {
```
tfenne, line 68: Thank you!

src/main/scala/com/fulcrumgenomics/commons/util/StringUtil.scala
```diff
@@ -154,4 +154,29 @@ object StringUtil {
         throw new IllegalArgumentException("columnIt failed.  Did you forget to input an NxM table?")
     }
   }
+
+  /** Splits line on delimiter, placing up to `arr.length` values into `arr` and returning the number of values. */
```
tfenne: Can you please add a line to the docs stating 

>Multiple consecutive delimeters are treated as individual delimiters, delimiting empty fields.  E.g. with `deliminter=','` the string `a,b,,,c` will return ["a", "b", "", "", "c"]`.

This is important and we rely on it in multiple places, but other `split` operations (e.g. on String`) behave differently.

src/test/scala/com/fulcrumgenomics/commons/CommonsDefTest.scala
```diff
@@ -128,6 +130,51 @@ class CommonsDefTest extends UnitSpec {
     }
   }
 
+  "CommonsDef.maxN" should "return the maximum N elements" in {
+    an[Exception] should be thrownBy Seq.empty[Int].maxN(0)
```
tfenne: If I can have a longestHomopolymer of length 0 in an empty string, why can't I have the top 0 things from an empty seq?

src/test/scala/com/fulcrumgenomics/commons/CommonsDefTest.scala
```diff
@@ -128,6 +130,51 @@ class CommonsDefTest extends UnitSpec {
...
+    Range.inclusive(start=1, end=seq.length*2).foreach { n => // tests asking both fewer and more than the # of elements
+      //seq.maxNBy(n, s => -s) should contain theSameElementsInOrderAs sorted.take(n)
+      seq.maxNBy(n, s => -s, distinct=true) should contain theSameElementsInOrderAs sorted.distinct.take(n)
+    }
+  }
+
+
+  // Comment this out to get performance testing
+  /**
+  "CommonsDef.maxN" should "should run quicly" in {
```
tfenne: quic[k]ly

src/test/scala/com/fulcrumgenomics/commons/util/StringUtilTest.scala
tfenne: 😄

src/main/scala/com/fulcrumgenomics/commons/async/Async.scala
nh13: So I have added more doc, since obviously it is unclear.

This class is meant to be mixed in to:
1. Build one or more threads (or runnables now) that each execute the class' `run()` method.
2. Gives a method `checkAndRethrow` that can check if the thread or runnable has thrown an exception, and if so, rethrows it in the class.   This may be useful to call when the subsequent code relies on the threads not have an exception.
3. Gives a method `tryAndModifyInterruptedException` that allows blocking tasks to be done such that any interrupted exceptions are turned into an unchecked (runtime) exception.  This is useful for blocking code where when interrupted, it is unrecoverable.

 If an implementing class uses/builds one thread, it is just performing asynchronous processing, while if it builds multiple threads, it is doing asynchronous _parallel_ processing.

src/main/scala/com/fulcrumgenomics/commons/async/Async.scala
nh13: I added a method `buildRunnable` for this use case.

src/main/scala/com/fulcrumgenomics/commons/async/Async.scala
nh13: This should be used internally by the classes that extend this trait.

src/main/scala/com/fulcrumgenomics/commons/async/Async.scala
nh13: Since I am using an `ArrayBlockingQueue`, having it be a large number would be bad.

src/main/scala/com/fulcrumgenomics/commons/async/Async.scala
nh13: Also, it may not always thrown something, so it needs to be `Unit`

src/main/scala/com/fulcrumgenomics/commons/async/Async.scala
nh13: done here and throughout

src/main/scala/com/fulcrumgenomics/commons/async/Async.scala
nh13: Since the parent trait can spawn one or more threads, it isn't true that when thread completes, all work is done.

src/main/scala/com/fulcrumgenomics/commons/async/Async.scala
nh13: I am keeping it as is for simplicity.

src/main/scala/com/fulcrumgenomics/commons/async/Async.scala
nh13: isDone doesn't join it.  I am not sure what you're asking.

src/main/scala/com/fulcrumgenomics/commons/async/Async.scala
nh13: Granted I adapted this from some of your Java code, I think the idea is that we want to detect any exception in the thread as early as possible, and in this case, check after every blocking call.

src/main/scala/com/fulcrumgenomics/commons/async/Async.scala
nh13: Done

src/main/scala/com/fulcrumgenomics/commons/async/Async.scala
nh13: Reminder: I adapted something you wrote: https://github.com/samtools/htsjdk/blob/master/src/main/java/htsjdk/samtools/util/AbstractAsyncWriter.java

;)

src/main/scala/com/fulcrumgenomics/commons/util/StringUtil.scala
nh13: Done

src/test/scala/com/fulcrumgenomics/commons/CommonsDefTest.scala
nh13: Great point, fixed.

nh13 (commented):
@tfenne lets chat about the async stuff, as I don't think I was clear about what the `AsyncThread` (now `AsyncThreadBuilder`) trait is trying to facilitate

nh13:
@tfenne I have refactored the `Async` classes based on our previous discussions.  I think it is much better.  There are certainly more things we can add to `AsyncRunnable`, but I am trying to keep it small(ish) for now.

src/main/scala/com/fulcrumgenomics/commons/async/AsyncIterator.scala
```diff
@@ -0,0 +1,94 @@
...
+
+  protected def execute(): Unit = this.source.foreach(this.queue.put)
+
+  /** Returns true if there exists more elements, false otherwise */
+  def hasNext(): Boolean = {
+    checkAndRaise()
+
+    // If the queue is empty but the underlying thread is not done, then there may or may not be new
+    // items that will be put in the queue.
+    while (this.queue.isEmpty && !this.done) {
```
tfenne: I think this could be done as the following if you cached the next item like in a buffered iterator:

```scala
var item = null
while (item == null && !this.done) item = this.queue.poll(50, TimeUnit.Milliseconds)
...
```

This has the advantage that instead of always waiting 50ms when the queue is empty and the thread is working, it will wait _at most_ 50ms, but should return much sooner if an item is inserted into the queue.

src/main/scala/com/fulcrumgenomics/commons/async/AsyncIterator.scala
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    // At this point, either there are items in the queue, or the underlying thread has consumed the source
+    if (!this.queue.isEmpty) true
+    else { // the queue is empty, and the source has been consumed
+      // wait for the execute method to complete, regardless of it an exception was thrown
+      awaitDone()
+      // check for an exception
+      checkAndRaise()
+      !this.queue.isEmpty
```
tfenne: Shouldn't the queue always be empty since the thread is "done" and you checked this on line 73?  If so I think it would be clearer to return `false` here directly.

src/main/scala/com/fulcrumgenomics/commons/async/AsyncIterator.scala
```diff
@@ -0,0 +1,94 @@
...
+    }
+  }
+
+  /** Gets the next item. */
+  def next(): T = {
+    checkAndRaise()
+    if (this.hasNext()) {
+      val item = tryAndModifyInterruptedException("Interrupted waiting on taking from the queue.") { this.queue.take() }
```
tfenne: You should be able to use `this.queue.remove()` here which is a non-blocking method that throws an exception if no element is available, but then you don't need to worry about `InterruptedException`s.

src/main/scala/com/fulcrumgenomics/commons/async/AsyncRunnable.scala
```diff
@@ -0,0 +1,168 @@
...
+
+package com.fulcrumgenomics.commons.async
+
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.atomic.{AtomicInteger, AtomicReference}
+
+import scala.concurrent.{Awaitable, ExecutionContext, Future}
+
+object AsyncRunnable {
+
```
tfenne: Bonus newline!

src/main/scala/com/fulcrumgenomics/commons/async/AsyncRunnable.scala
```diff
@@ -0,0 +1,168 @@
...
+  /** This signals that the run method has started. */
+  private val startedLatch: CountDownLatch = new CountDownLatch(1)
+
+  /** This signals that the run method has completed, even if an exception occurred. This is needed so the close method
+    * can be sure that the run method is no longer executing. */
+  private val doneLatch: CountDownLatch = new CountDownLatch(1)
+
+  final def run(): Unit = {
+    val outer = this
```
tfenne: FYI, you never need to do this.  Here, I think you can just use `this`.  But even in the case of nested classes you can do the following:

```scala
class Foo {
    def close(): Unit = println("Foo closing!")
    class Bar {
        def close(): Unit = Foo.this.close()
    }
}
```

src/main/scala/com/fulcrumgenomics/commons/async/AsyncRunnable.scala
```diff
@@ -0,0 +1,168 @@
...
+  * that are not part of the `execute()` method.  This method will throw a [[RuntimeException]] if interrupted while
+  * waiting.
+  */
+trait AsyncRunnable extends Runnable {
+
+  /** A reference to an exception thrown during an asynchronous operation. */
+  private[AsyncRunnable] val _throwable: AtomicReference[Throwable] = new AtomicReference(null)
+
+  /** This signals that the run method has started. */
+  private val startedLatch: CountDownLatch = new CountDownLatch(1)
```
tfenne, line 65: I'm curious if there's a reason to prefer this over say `AtomicBoolean`?

src/main/scala/com/fulcrumgenomics/commons/async/AsyncRunnable.scala
```diff
@@ -0,0 +1,168 @@
...
+
+  /** Returns a throwable if an exception occurred in the [[run()]] method, None otherwise. */
+  def throwable: Option[Throwable] = Option(this._throwable.get())
+
+  /** Waits for the [[run()]] method to start. */
+  def awaitStart(): Unit = this.startedLatch.await()
+
+  /** Returns an [[Awaitable]] that completes when the [[run()]] method has started.  Returns the throwable if an
+    * exception was encountered, [[None]] otherwise. */
+  def startAwaitable()(implicit ec: ExecutionContext) : Awaitable[Unit] = Future { this.awaitStart() }
```
tfenne: I'm having trouble coming up with a better name, but to me this sounds imperative like "start the awaitable!".

src/main/scala/com/fulcrumgenomics/commons/async/AsyncRunnable.scala
```diff
@@ -0,0 +1,168 @@
...
+    val threadName = name.getOrElse { AsyncRunnable.nextName(this) }
+    new Thread(this, threadName)
+  }
+
+  /** Starts this [[Runnable]] in a daemon thread.
+    *
+    * @param name optionally the name of the thread, otherwise a name is created based on the class name and the number
+    *             of threads already created.
+    */
+  final def start(name: Option[String] = None): this.type = {
```
tfenne: How about making `daemon: Boolean = true` a parameter?

src/main/scala/com/fulcrumgenomics/commons/async/AsyncRunnable.scala
```diff
@@ -0,0 +1,168 @@
...
+  /** Returns an [[Awaitable]] that completes when the [[run()]] method has completed.  Returns the throwable if an
+    * exception was encountered, [[None]] otherwise. */
+  def doneAwaitable()(implicit ec: ExecutionContext) : Awaitable[Option[Throwable]] = Future { this.awaitDone(); Option(this._throwable.get()) }
+
+  /** Creates a new thread wrapping this runnable; the thread is not started.
+    *
+    * @param name optionally the name of the thread, otherwise a name is created based on the class name and the number
+    *             of threads already created.
+    */
+  final def thread(name: Option[String] = None): Thread = {
```
tfenne: I'm curious why this is final and public?  I.e. I can't override it, and should I really be calling it directly?

src/main/scala/com/fulcrumgenomics/commons/async/AsyncRunnable.scala
tfenne: Also, how about making `name` an instance method that defaults to `AsyncRunnable.name(this)`?  That way naming is also encapsulated in the class and not delegated to the user (who can still extend/override if they need to)?

src/main/scala/com/fulcrumgenomics/commons/async/AsyncRunnable.scala
```diff
@@ -0,0 +1,168 @@
...
+  /** Checks to see if an exception has been raised by an asynchronous thread and if so rethrows it as an [[Error]]
+    * or [[RuntimeException]] as appropriate.  Use this method before code that assumes the threads have not encountered
+    * an exception.
+    */
+  protected final def checkAndRaise(): Unit = {
+    _throwable.getAndSet(null) match {
+      case null                 => Unit
+      case e: Error             => throw e
+      case ex: RuntimeException => throw ex
+      case t                    => throw new RuntimeException(t)
```
tfenne: I'm curious why this is necessary?  All exceptions are unchecked in scala...

src/main/scala/com/fulcrumgenomics/commons/async/AsyncSink.scala
```diff
@@ -0,0 +1,120 @@
...
+    * @param source the optional source to close when this sink is closed.
+    * @tparam T the type of object to process
+    */
+  def apply[T](sink: T => Unit, bufferSize: Option[Int] = None, source: Option[{ def close(): Unit }] = None): AsyncSink[T] = {
+    new AsyncSink[T](sink, bufferSize, source).start()
+  }
+}
+
+
+/** An asynchronous wrapper for processing items of type [[T]] and process them in a separate thread using
```
tfenne: Sentence doesn't scan.

src/main/scala/com/fulcrumgenomics/commons/async/AsyncWriter.scala
```diff
@@ -0,0 +1,52 @@
...
+  }
+}
+
+/** An asynchronous wrapper around a [[Writer]] class.
+  *
+  * @param writer the writer to wrap
+  * @param bufferSize the number of elements to buffer before blocking when writing the elements, or None if unbounded
+  * @tparam T the type of object to write
+  */
+class AsyncWriter[T](val writer: Writer[T], bufferSize: Option[Int] = None)
```
tfenne, line 48: Very tidy!

tfenne (approved):

src/main/scala/com/fulcrumgenomics/commons/async/AsyncIterator.scala
nh13: Done

src/main/scala/com/fulcrumgenomics/commons/async/AsyncIterator.scala
nh13: Done

src/main/scala/com/fulcrumgenomics/commons/async/AsyncIterator.scala
nh13: Done

src/main/scala/com/fulcrumgenomics/commons/async/AsyncRunnable.scala
nh13: NO!

src/main/scala/com/fulcrumgenomics/commons/async/AsyncRunnable.scala
nh13: I know it's private, but `CountDownLatch` can only be decremented, whereas its up to you (me) to set `AtomicBoolean` from `false` to `true` and not back again.  I don't trust myself not to screw it up.

src/main/scala/com/fulcrumgenomics/commons/async/AsyncRunnable.scala
nh13: Cool!

src/main/scala/com/fulcrumgenomics/commons/async/AsyncRunnable.scala
nh13: How about `uponStart` and `uponDone`?

src/main/scala/com/fulcrumgenomics/commons/async/AsyncRunnable.scala
nh13: This method gives you control of what to do with the thread without it being started.  It's final, since I don't want folks messing with this method, as it has some logic that isn't public about setting the thread name.  We can loosen the `final` later if we want to. I  was just erring on the side of saying "hey it's final because there's some stuff in here that you can't access if you override it".  ¯\_(ツ)_/¯

src/main/scala/com/fulcrumgenomics/commons/async/AsyncRunnable.scala
nh13: I think name would have to be a `val` since if it is a `def` then each call to name would return a different value based on how `AsyncRunnable.nextName` works.  Done.

src/main/scala/com/fulcrumgenomics/commons/async/AsyncRunnable.scala
nh13: You're right, so it can be:
```
    _throwable.getAndSet(null) match {
      case null           => Unit
      case thr: Throwable => throw thr
    }
```

nh13:
@tfenne all done!

nh13:
@tfenne it also looks like there's a failure in `AsyncWriterPool`: https://travis-ci.org/fulcrumgenomics/commons/builds/421919739

Care to take a look?

PR #35, opened 2018-08-18
A more general pooling mechanism for asynchronous writers.

tfenne:
@nh13 Here's my take on how to pool threads for many writers.  I'm curious both how you like it and how it performs for DemuxFastqs relative to `AsyncMultiWriter`.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/commons/async/AsyncWriterPool.scala
+++ b/src/main/scala/com/fulcrumgenomics/commons/async/AsyncWriterPool.scala
@@ -1,0 +1,175 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2018 Fulcrum Genomics
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ */
+package com.fulcrumgenomics.commons.async
+
+import java.io.Closeable
+import java.util.concurrent._
+import java.util.concurrent.atomic.{AtomicBoolean, AtomicReference}
+
+import com.fulcrumgenomics.commons.CommonsDef._
+import com.fulcrumgenomics.commons.io.Writer
+
+/**
+  * A shared thread pool for making asynchronous [[Writer]]s without allocating a thread per writer.
+  * The pool is created with a fixed number of threads which are used to handle write operations for
+  * all writers associated to the pool.
+  *
+  * Writers are added to the pool using the [[pool()]] method, which returns a [[PooledWriter]] that
+  * can be written to in place of the original [[Writer]].  [[Writer]]s of different types can be mixed
+  * within the same pool, and each writer can specify either an unbounded queue or a fixed queue of
+  * independent size when being pooled.
+  *
+  * [[Writer]] can be added to the pool at any time (except aftering calling [[close()]] on the pool.
+  * Writers can be closed individually (and removed from the pool) using either their [[PooledWriter.close()]]
+  * method which runs synchronously on the calling thread, or by using [[PooledWriter.closeAsync()]]
+  * which will execute the close asynchronously and return a [[Future]] upon which `get()` must be called
+  * to re-throw any exceptions that were thrown during the close operation.
+  *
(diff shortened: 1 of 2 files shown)


src/test/scala/com/fulcrumgenomics/commons/async/AsyncWriterPoolTest.scala
```diff
@@ -0,0 +1,66 @@
...
+
+    for (queueSize <- Seq(None, Some(10), Some(100))) {
+      Seq((2, 4), (5, 3), (3, 3)).foreach { case (numWriters, numThreads) =>
+        val qs = queueSize match {
+          case None    => "unbounded queues"
+          case Some(l) => s"$l-length queues"
+        }
+
+        it should s"write with $numWriters writers, $numThreads threads and $qs" in {
+          val pool    = new AsyncWriterPool(threads=numThreads)
```
nh13, line 52: `qs` isn't being used

src/test/scala/com/fulcrumgenomics/commons/async/AsyncWriterPoolTest.scala
```diff
@@ -0,0 +1,66 @@
...
+            writer.write(string)
+          }
+
+          pool.close()
+          writers.forall(_.closed) shouldBe true
+          writers.foreach(w => w.items should contain theSameElementsInOrderAs strings)
+        }
+      }
+    }
+  }
```
nh13, line 66: Could you add just a few more tests?

- have a writer throw an exception, and see that you it is propagated in each of the `checkAndRaise` places?
- deadlock if the buffer is too small (optional)

src/main/scala/com/fulcrumgenomics/commons/async/AsyncWriterPool.scala
```diff
@@ -0,0 +1,175 @@
...
+  private val executor   = Executors.newFixedThreadPool(threads)
+  private val writers    = new CopyOnWriteArrayList[PooledWriter[_]]()
+  private val poolClosed = new AtomicBoolean(false)
+
+  /**
+    * Wraps a [[Writer[A]]] in the context of an [[AsyncWriterPool]] to provide asynchronous operations.
+    * Calls to [[write()]] and related methods place items into the queue, and if necessary, schedule
+    * the queue for draining using the [[executor]] in the enclosing [[AsyncWriterPool]].
+    */
+  class PooledWriter[A] private[AsyncWriterPool] (private val writer: Writer[A], private val queue: BlockingQueue[A]) extends Writer[A] {
```
nh13: Now I am wondering if this should even be on `Writer`s.   Should we generalize it so we have `AsyncMultiSink`, then have a second class that accepts writers, just like `AsyncSink/AsyncWriter`?  I think that would be nicer.

nh13 (changes requested):
I am happy to help on the refactor and/or test writing.

src/main/scala/com/fulcrumgenomics/commons/async/AsyncWriterPool.scala
```diff
@@ -0,0 +1,188 @@
...
+      checkAndRaise()
+      writerCloseRequested.set(true)
+      executor.submit(new Callable[Unit] { override def call(): Unit = PooledWriter.this.close() })
+    }
+  }
+
+  /** A runnable that attempts to drain items from all queues to their writers. */
+  private class Drainer extends Runnable {
+    override def run(): Unit = try {
+      while (!poolClosed.get()) {
```
nh13, line 132: Should we sleep each iteration? `Drainer` is created in the constructor is so is runnable all the time.

src/main/scala/com/fulcrumgenomics/commons/async/AsyncWriterPool.scala
```diff
@@ -0,0 +1,188 @@
...
+                                                  private val queue: BlockingQueue[A]) extends Writer[A] {
+    /** A reference to an exception thrown during an asynchronous operation. */
+    private[AsyncWriterPool] val throwable: AtomicReference[Throwable] = new AtomicReference(null)
+
+    /** True if this writer is enqueued to be drained, false otherwise. */
+    private[AsyncWriterPool] val draining: AtomicBoolean = new AtomicBoolean(false)
+
+    /** Drains the queue to the underlying writer. Returns the count of items drained. */
+    private[AsyncWriterPool] def drain(): Int = {
+      this.synchronized {
```
nh13, line 78: move this to the previous line?

src/main/scala/com/fulcrumgenomics/commons/async/AsyncWriterPool.scala
```diff
@@ -0,0 +1,188 @@
...
+      queue.put(item)
+    }
+
+    /** Closes the writer _synchronously_ on the current thread. */
+    override def close(): Unit = {
+      checkAndRaise()
+      writerCloseRequested.set(true)
+      AsyncWriterPool.this.writers.remove(this)
+      if (!writerClosed.getAndSet(true)) {
+        drain()
```
nh13, line 116: Should we check that this writer isn't draining before calling `drain()`?  The `Drainer`s you created could be draining.  I suppose that because `drain()` is synchronized it shouldn't matter.

src/main/scala/com/fulcrumgenomics/commons/async/AsyncWriterPool.scala
```diff
@@ -0,0 +1,188 @@
...
+  private val executor   = Executors.newFixedThreadPool(threads)
+  private val writers    = new CopyOnWriteArrayList[PooledWriter[_]]()
+  private val poolClosed = new AtomicBoolean(false)
+  Range(0, threads).foreach(_ => executor.submit(new Drainer))
+
+  /**
+    * Wraps a [[Writer[A]]] in the context of an [[AsyncWriterPool]] to provide asynchronous operations.
+    * Calls to [[write()]] and related methods place items into the queue, and if necessary, schedule
+    * the queue for draining using the [[executor]] in the enclosing [[AsyncWriterPool]].
+    */
```
nh13, line 67: Perhaps some notes for our future selves is appropriate here.  For example, how we spawn a `Drainer` per thread to actively drain the queues while the pool has writers.  Then how there is one final `drain()` when you close a writer.

src/main/scala/com/fulcrumgenomics/commons/async/AsyncWriterPool.scala
nh13: You likely want to wrap it in `blocking{}` if you sleep.

nh13 (approved):
A few nitpick implementation details.  Can you take a look?

nh13:
thanks @tfenne !

PR #36, opened 2018-11-03
Workaround for Files.newInputStream() returning a stream that's incom…

tfenne:
…patible with stdin.


nh13 (approved):

nh13:
@tfenne FYI, the build failure is the one I see every once in a while with the `AsyncWriterPool` I mentioned previously

PR #42, opened 2018-12-24
Make the parser faster when accessing field value as Strings.

tfenne:
@nh13 I really like our `DelimitedDataParser` but all the reflection makes it really slow when dealing with large files (100,000s or millions of rows).  I have a tool I'm working on that is processing files of many millions of rows.  I'm accessing all the values as Strings, and switching away from calling the reflection code makes is over 10X faster overall!! 

Totally open to changing the names of the methods if you have a better suggestion.  I was hoping I could implement it as just `def apply(column: String): String`.  That does compile, but then when you try to use it you get compile failures because it's ambiguous :(

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/commons/util/DelimitedDataParser.scala
+++ b/src/main/scala/com/fulcrumgenomics/commons/util/DelimitedDataParser.scala
@@ -36,7 +36,7 @@ import scala.reflect.runtime.{universe => ru}
   */
 class Row private[util] (private val headerIndices: Map[String,Int], private val fields: Array[String], val trim: Boolean) {
   /* Internal method to pull a value out of the field array by index and trim it if requested. */
-  private def value(index: Int) = {
+  private def value(index: Int): String = {
     if (index > fields.length-1) throw new IndexOutOfBoundsException(s"Invalid column index supplied: ${index}")
     val string = fields(index)
     if (trim) string.trim else string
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/commons/util/DelimitedDataParser.scala
```diff
@@ -51,6 +51,12 @@ class Row private[util] (private val headerIndices: Map[String,Int], private val
   /** Fetches a value of the desired type by column name. */
   def apply[A](column: String)(implicit tag: ru.TypeTag[A]): A = apply(headerIndices(column))
 
+  /** Non-type-parameterized apply method that returns the column value as a String. */
+  def string(column: String): String = value(headerIndices(column))
+
+  /** Non-type-parameterized apply method that returns the column value as a String. */
+  def string(index: Int): String = value(index)
```
nh13, line 58: I like `string` but it could also be named `raw`?

nh13: I also wonder if you could check the typetag in apply to see if it is string, and then just short-circuit to return value, and if that would speed it up.

nh13 (approved):

src/main/scala/com/fulcrumgenomics/commons/util/DelimitedDataParser.scala
tfenne: I would really like to do the latter, but I couldn't find a way to check the type tag that was a) significantly faster than the existing implementation or b) anywhere close to as fast as a non-type-var'd method.

I'll stick with `string` if you don't mind since it still returns the trimmed string, so I think `raw` could be a touch misleading.

PR #44, opened 2019-03-25
Add FilenameSuffix type alias to CommonsDef

clintval:
I find the type aliases in `CommonsDef.scala` very useful but would love to see a `FilenameSuffix` type alias since I do a lot of work with building paths using suffix extension.

```scala
val bamSamplePrefix: FilenamePrefix  = "sample.library"
val bamGroupedSuffix: FilenameSuffix = ".grouped"
val bamSuffix: FilenameSuffix        = ".bam"

val bamPath: PathToBam = PathUtil.pathTo(bamSamplePrefix + bamGroupedSuffix + bamSuffix)
```

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/commons/CommonsDef.scala
+++ b/src/main/scala/com/fulcrumgenomics/commons/CommonsDef.scala
@@ -25,15 +25,13 @@
 package com.fulcrumgenomics.commons
 
 import java.io.Closeable
-import java.util.Comparator
+import java.util.concurrent.ForkJoinPool
 
 import com.fulcrumgenomics.commons.collection.BetterBufferedIterator
 import com.fulcrumgenomics.commons.util.Logger
 
-import scala.collection.mutable.ListBuffer
-import scala.collection.{LinearSeq, LinearSeqOptimized, Parallelizable, SeqLike}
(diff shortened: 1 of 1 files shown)


src/main/scala/com/fulcrumgenomics/commons/CommonsDef.scala
```diff
@@ -25,15 +25,13 @@
 package com.fulcrumgenomics.commons
 
 import java.io.Closeable
-import java.util.Comparator
+import java.util.concurrent.ForkJoinPool
```
clintval: Imports were optimized (with IntelliJ) and one deprecation warning was solved with this import:

Deprecated import (alias):

https://www.scala-lang.org/api/current/scala/concurrent/forkjoin/package$$ForkJoinPool$.html

tfenne: Unfortunately we can't do this while we're still cross-building for 2.11 and 2.12 - it breaks the 2.11 build.

tfenne:
Happy to merge this @clintval once you revert the change to the ForkJoinPool import.

clintval:
Woops! Back to you.

PR #45, opened 2019-05-07
DelimitedDataParser to accept a sequence of header names

clintval:
It's common for me to parse files without headers and I would like to supply my own.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/commons/util/DelimitedDataParser.scala
+++ b/src/main/scala/com/fulcrumgenomics/commons/util/DelimitedDataParser.scala
@@ -86,7 +86,7 @@ object DelimitedDataParser {
 
   /** Constructs a DelimitedDataParser for a path. */
   def apply(path: FilePath, delimiter: Char, header: Seq[String]): DelimitedDataParser =
-    new DelimitedDataParser(Io.toSource(path).getLines(), delimiter=delimiter, header=header)
+    apply(Io.toSource(path).getLines(), delimiter=delimiter, header=header)
 
   /** Constructs a DelimitedDataParser for a sequence of lines. */
   def apply(lines: TraversableOnce[String], delimiter: Char): DelimitedDataParser =
@@ -104,7 +104,7 @@ object DelimitedDataParser {
   * @param delimiter the delimiter between columns
   * @param ignoreBlankLines whether blank lines should be ignored
   * @param trimFields whether individual fields should have their String values trimmed
-  * @param header the header names for the columns of delimited data
+  * @param header the header names for the columns of delimited data. If empty, the first line should contain the header names.
   */
 class DelimitedDataParser(lines: TraversableOnce[String],
                           val delimiter: Char,
--- a/src/main/scala/com/fulcrumgenomics/commons/util/DelimitedDataParser.scala
+++ b/src/main/scala/com/fulcrumgenomics/commons/util/DelimitedDataParser.scala
@@ -85,8 +85,16 @@ object DelimitedDataParser {
     new DelimitedDataParser(Io.toSource(path).getLines(), delimiter=delimiter)
 
   /** Constructs a DelimitedDataParser for a path. */
+  def apply(path: FilePath, delimiter: Char, header: Seq[String]): DelimitedDataParser =
+    new DelimitedDataParser(Io.toSource(path).getLines(), delimiter=delimiter, header=header)
+
+  /** Constructs a DelimitedDataParser for a path. */
   def apply(lines: TraversableOnce[String], delimiter: Char): DelimitedDataParser =
     new DelimitedDataParser(lines, delimiter=delimiter)
+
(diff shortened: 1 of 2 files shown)


src/test/scala/com/fulcrumgenomics/commons/util/DelimitedDataParserTest.scala
```diff
@@ -25,10 +25,16 @@
 package com.fulcrumgenomics.commons.util
 
 import com.fulcrumgenomics.commons.CommonsDef._
-
+import com.fulcrumgenomics.commons.io.Io
```
nh13, line 28: newline after

src/test/scala/com/fulcrumgenomics/commons/util/DelimitedDataParserTest.scala
```diff
@@ -25,10 +25,16 @@
...
 /** Tests for the delimited data parser. */
 class DelimitedDataParserTest extends UnitSpec {
-  def csv(lines: Seq[String], trim: Boolean=true) = new DelimitedDataParser(lines, delimiter=',', trimFields=trim)
+  private lazy val csvFile: FilePath = {
```
nh13: No need to be lazy since the tests will definitely be run.

src/main/scala/com/fulcrumgenomics/commons/util/DelimitedDataParser.scala
```diff
@@ -84,9 +84,17 @@ object DelimitedDataParser {
   def apply(path: FilePath, delimiter: Char): DelimitedDataParser =
     new DelimitedDataParser(Io.toSource(path).getLines(), delimiter=delimiter)
 
+  /** Constructs a DelimitedDataParser for a path. */
+  def apply(path: FilePath, delimiter: Char, header: Seq[String]): DelimitedDataParser =
+    new DelimitedDataParser(Io.toSource(path).getLines(), delimiter=delimiter, header=header)
```
nh13: It's probably good to re-route the apply on 84-85 to use this now one with `header=Seq.empty`. Ditto below.

nh13 (changes requested):
Really minor stuff, thanks for adding this.

src/main/scala/com/fulcrumgenomics/commons/util/DelimitedDataParser.scala
clintval: Scala will not allow me to have multiple overloaded alternatives of the same method with default arguments defined:

https://stackoverflow.com/a/10138519/3727678

https://gist.github.com/lossyrob/63f0d40506ff8b4c3b22

nh13: I am not asking to have default arguments in the function signature.  I am asking to have the `apply` method _without_ the `header` parameter call the `apply` method _with_ the `header` parameter, so if we change the call to the latter, the former does not need updating. For example:

```scala
  /** Constructs a DelimitedDataParser for a path. */
  def apply(path: FilePath, delimiter: Char, header: Seq[String]): DelimitedDataParser =
    new DelimitedDataParser(Io.toSource(path).getLines(), delimiter=delimiter, header=header)

  /** Constructs a DelimitedDataParser for a path. */
  def apply(lines: TraversableOnce[String], delimiter: Char): DelimitedDataParser =
    this.apply(lines, delimiter=delimiter, header=Seq.empty)
```

clintval: oh-yea-duh

clintval:
Back to you, thanks!

src/main/scala/com/fulcrumgenomics/commons/util/DelimitedDataParser.scala
nh13: 64CC31C0-60DF-4ABE-BE37-622B3298625F

nh13: @clintval I think L89 can be:
```suggestion
    apply(Io.toSource(path).getLines(), delimiter=delimiter, header=header)
```

src/main/scala/com/fulcrumgenomics/commons/util/DelimitedDataParser.scala
```diff
@@ -96,17 +104,20 @@ object DelimitedDataParser {
   * @param delimiter the delimiter between columns
   * @param ignoreBlankLines whether blank lines should be ignored
   * @param trimFields whether individual fields should have their String values trimmed
+  * @param header the header names for the columns of delimited data
```
nh13: Can you add
```suggestion
  * @param header the header names for the columns of delimited data.  If empty, the first line should contain the header names.
```

nh13 (changes requested):
@clintval one more possible `apply` and one comment I missed.  Thanks!

clintval:
Thx!

PR #46, opened 2019-05-17
Add the sealedTraitOptions method to return the direct sub-classes of…

nh13:
… a trait

These sub-classes must be case objects.

Will be used here: https://github.com/fulcrumgenomics/sopt/pull/24

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/commons/reflect/ReflectionUtil.scala
+++ b/src/main/scala/com/fulcrumgenomics/commons/reflect/ReflectionUtil.scala
@@ -29,7 +29,7 @@ import java.nio.file.Path
 import com.fulcrumgenomics.commons.io.PathUtil
 import com.fulcrumgenomics.commons.CommonsDef._
 
-import scala.annotation.ClassfileAnnotation
+import scala.annotation.{ClassfileAnnotation, tailrec}
 import scala.collection.mutable
 import scala.reflect._
 import scala.reflect.runtime.universe._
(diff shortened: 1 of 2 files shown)


src/test/scala/com/fulcrumgenomics/commons/reflect/ReflectionUtilTest.scala
```diff
@@ -61,6 +61,42 @@ object ReflectionUtilTest {
...
+
+/** Sealed-trait hierarchy where `findValues` returns the values in a different order than the case objects they were declared */
+sealed trait CompanionWithFindValues
+case object ACompanionWithFindValues extends CompanionWithFindValues
+case object BCompanionWithFindValues extends CompanionWithFindValues
+case object CCompanionWithFindValues extends CompanionWithFindValues
+object CompanionWithFindValues {
+  def findValues: Seq[CompanionWithFindValues] = Seq(BCompanionWithFindValues, ACompanionWithFindValues, CCompanionWithFindValues)
+}
+
```
tfenne, line 94: Should we also test cases where the case objects are declared inside the companion object?  That's the usual pattern for enumeratum enums, so it would be good to make sure it works for those.

src/main/scala/com/fulcrumgenomics/commons/reflect/ReflectionUtil.scala
```diff
@@ -434,4 +433,48 @@ object ReflectionUtil {
...
+          .map {
+            case Success(_values) => _values
+            case Failure(_)       => Seq.empty[Any]
+          }
+          .find(_.nonEmpty)
+    }
+
+    companionValues.map { values => copy(values.toArray, clazz).toSeq }.getOrElse {
+      // Default to the direct sub-classes of the trait
+      symbol.knownDirectSubclasses.map { subSymbol =>
```
tfenne: Would it be hard to support indirect subclasses?  Ideally it would find all non-abstract subclasses I think.  I have at least one case somewhere where I have a sealed trait, with sub-traits and then a bunch of case objects extending the sub-traits.

tfenne (commented):
A couple of comments ... let me know what you think.

src/test/scala/com/fulcrumgenomics/commons/reflect/ReflectionUtilTest.scala
nh13: Good idea, done.

src/main/scala/com/fulcrumgenomics/commons/reflect/ReflectionUtil.scala
nh13: Sounds good.

PR #47, opened 2019-05-24
Change paths in coverage tests for sbt layout

nvander1:
@nh13 RE: https://github.com/lihaoyi/mill/issues/613
Glad to see you're using Mill + Scoverage!

## Part 1 (Fixing test running under scoverage)
For your first part of the issue (the tests that weren't running). We need to modify where mill is looking for its source files in the test. To get it to find the right files, we follow how the mill `SbtModule` adjusts its search: https://github.com/lihaoyi/mill/blob/master/scalalib/src/MiscModule.scala#L66-L77.

We can verify that the paths are the same with the following commands:
```
mill show commons.test.sources
mill show commons.coverage.sources

mill show commons.test.resources
mill show commons.coverage.resources
```

And you'll see that the following command completes successfully:
```
mill commons.coverage
```

## Part 2 (Fixing html report exception)

I've identified the bug causing the htmlReport exception and have implemented a fix to merge into mill: https://github.com/lihaoyi/mill/pull/615.

Once the fix is merged in and a new version is released, you will be able to run the htmlReport command successfully.

## Note about mixed Java and Scala sources
Scoverage should be able to track the statement coverage for your Scala sources, but I don't think it will
create reports for Java sources.

See the following for more information about this:
- https://github.com/scoverage/gradle-scoverage/issues/62
- http://buransky.com/scala/the-best-code-coverage-for-scala/

Perhaps @RadoBuransky can comment more on the second link since they wrote the article! :)


nh13:
@nvander1 this is great info!  For (1), since we are moving to mill (see #48), I am changing the project layout, which makes things a lot easier.  For (2), thank-you for the fix (tested and it works).  I think it is fine that we only measure coverage on our scala code, as we aim to minimize the amount of Java code.  Unfortunately, sometimes Java is necessary.

PR #48, opened 2019-05-27
Changing to the mill build system

nh13:
Note: this isn't ready until:
- [ ] mill is updated to output XML scoverage reports (https://github.com/lihaoyi/mill/pull/618)
- [ ] travis-ci is able to pass
- [ ] we have tested mill with publishing to sonatype


build.sc
```diff
@@ -0,0 +1,56 @@
...
+
+import mill.scalalib.{PublishModule, Dep, DepSyntax, ScalaModule}
+import mill.scalalib.publish.{PomSettings, License, Developer, SCM}
+import ammonite.ops._
+
+import scala.sys.process.Process
+
+object commons extends ScalaModule with ScoverageModule with PublishModule {
+  def artifactName = "commons"
+  def gitHash = Process("git rev-parse --short HEAD").lineStream.head
```
jacarey: I think  you can avoid using Process and instead use ammonite os.proc

jacarey (approved):

.travis.yml
```diff
@@ -1,24 +1,24 @@
...
+
 before_install:
-- pip install --user codecov
-- sudo chmod +x /usr/local/bin/sbt
+  - pip install --user codecov
 after_success:
 - codecov
 - test "$TRAVIS_REPO_SLUG" == 'fulcrumgenomics/commons' && test "$TRAVIS_PULL_REQUEST"
-  == 'false' && test "$TRAVIS_BRANCH" == 'master' && sbt +publish
+  == 'false' && test "$TRAVIS_BRANCH" == 'master' && mill mill.scalalib.PublishModule/publishAll "${SONATYPE_USERNAME}:${SONATYPE_PASSWORD}" __.publishArtifacts --release true
```
tfenne, line 21: This is a little cumbersome.  Is there a way to put some of this line into the build file so that the command is closer to `mill publishAll`?  Thinking about when we're doing official releases and doing this by hand.

tfenne: Also, is this doing cross-builds?

README.md
```diff
@@ -3,7 +3,6 @@
 [![Code Review](https://api.codacy.com/project/badge/grade/52e1d786d9784c7192fae2f8e853fa34)](https://www.codacy.com/app/contact_32/commons)
 [![Maven Central](https://maven-badges.herokuapp.com/maven-central/com.fulcrumgenomics/commons_2.11/badge.svg)](https://maven-badges.herokuapp.com/maven-central/com.fulcrumgenomics/commons_2.11)
 [![Javadocs](http://javadoc.io/badge/com.fulcrumgenomics/commons_2.12.svg)](http://javadoc.io/doc/com.fulcrumgenomics/commons_2.12)
-[![Dependency Status](https://www.versioneye.com/user/projects/56b2d2d593b95a003c714340/badge.svg)](https://www.versioneye.com/user/projects/56b2d2d593b95a003c714340#dialog_dependency_badge)
```
tfenne, line 6: Could we also add a little to the README about how to build the project, so how to get the minimum version of mill and what the command is to compile/test/build jar? Perhaps how to publish locally?

build.sc
```diff
@@ -0,0 +1,56 @@
...
+
+object commons extends ScalaModule with ScoverageModule with PublishModule {
+  def artifactName = "commons"
+  def gitHash = os.proc("git", "rev-parse", "--short", "HEAD").call().out.trim
+  def publishVersion = s"0.8.0-${gitHash}-SNAPSHOT"
+  def scalaVersion = "2.12.8"
+  def scoverageVersion = "1.3.1"
+  def scalacOptions = Seq("-target:jvm-1.8", "-deprecation", "-unchecked")
+
+  // TODO: start year (2015)
```
tfenne, line 21: Can we resolve these TODOs?

README.md
tfenne: We should also think about how one gets setup in IntelliJ since IntelliJ doesn't (yet) understand mill build files.

build.sc
```diff
@@ -0,0 +1,56 @@
...
+import mill.contrib.scoverage.ScoverageModule
+
+import mill.scalalib.{PublishModule, Dep, DepSyntax, ScalaModule}
+import mill.scalalib.publish.{PomSettings, License, Developer, SCM}
+import ammonite.ops._
+
+import scala.sys.process.Process
+
+object commons extends ScalaModule with ScoverageModule with PublishModule {
+  def artifactName = "commons"
```
tfenne, line 14: Could we do something like `getClass.getSimpleName` here to standardize object == artifact name?  Or do we want to embed the version name here?

build.sc
```diff
@@ -0,0 +1,56 @@
...
+  def gitHash = os.proc("git", "rev-parse", "--short", "HEAD").call().out.trim
+  def publishVersion = s"0.8.0-${gitHash}-SNAPSHOT"
```
tfenne, line 16: I wonder if there's an easy way to pass custom params to a mill build such that we could define the version as `if (release) "0.8.0" else (s"....")`

build.sc
```diff
@@ -0,0 +1,56 @@
...
+  def scalaVersion = "2.12.8"
```
tfenne, line 17: So do we lose cross-building?  That seems less than ideal.

commons/test/src/com/fulcrumgenomics/commons/io/AsyncStreamSinkTest.scala
```diff
@@ -35,7 +35,7 @@ import scala.collection.mutable
   */
 class AsyncStreamSinkTest extends UnitSpec {
   "AsyncStreamSink" should "capture all the output" in {
-    val file = Paths.get("src/test/resources/com/fulcrumgenomics/commons/io/async-stream-sink-test.txt")
+    val file = Paths.get("commons/test/resources/com/fulcrumgenomics/commons/io/async-stream-sink-test.txt")
```
tfenne, line 38: I'm a bit surprised by this. Does mill require even for single-module builds that the main and test sources are within a module directory?

tfenne (changes requested):
A few questions and comments.

README.md
nh13: `mill mill.scalalib.GenIdea/idea; `

build.sc
nh13: I could use some help 👍

build.sc
nh13: Ah yes, I'll want to do add that support.  Mill has facilities to do this http://www.lihaoyi.com/mill/page/cross-builds.html

commons/test/src/com/fulcrumgenomics/commons/io/AsyncStreamSinkTest.scala
nh13: http://www.lihaoyi.com/mill/page/common-project-layouts.html

nh13:
@tfenne I'll address your feedback, but I added a few comments for your future reading.

PR #51, opened 2019-07-23
Thread local and a test for BiMap

nh13:


src/main/scala/com/fulcrumgenomics/commons/util/ThreadUtil.scala
```diff
@@ -0,0 +1,53 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ */
+
+package com.fulcrumgenomics.commons.util
+
+import com.fulcrumgenomics.commons.CommonsDef.javaIterableToIterator
+
+object ThreadUtil {
```
tfenne: I think we've been using the pattern we see in other scala libraries to name utils as the plural of the thing they are utils for, e.g. `Threads` vs. `ThreadUtil`.  Can we do that here please?

src/main/scala/com/fulcrumgenomics/commons/util/ThreadUtil.scala
```diff
@@ -0,0 +1,53 @@
...
+      val a = factory()
+      all.add(a)
+      a
+    }
+
+    /** Returns all the thread-local objects created by the given factory.
+      *
+      * Care should be taken accessing the iterator since objects may be in use by other threads.
+      */
+    def iterator: Iterator[A] = all.toIterator
```
tfenne: I think `toIterator` has been deprecated in 2.13.  Sorry!

tfenne (approved):
LGTM once the build is passing.

PR #54, opened 2019-07-25
Removes as many deprecations as I can while still cross-building

tfenne:
In the end I think it's easier to stop using symbols and instead just hit the methods directly, i.e. `foo shouldBe 'success` => `foo.isSuccess shouldBe true`.  I tried using some constant symbols, but it's ugly.  If you want to use, e.g. `Success` then that conflicts with the subclass of `Try` and causes other issues.

### DIFF
--- a/src/main/scala-2.12-/com/fulcrumgenomics/commons/Compat.scala
+++ b/src/main/scala-2.12-/com/fulcrumgenomics/commons/Compat.scala
@@ -2,11 +2,13 @@ package com.fulcrumgenomics.commons
 
 import scala.annotation.ClassfileAnnotation
 
-/** Trait that includes types used for cross-building compatibility; scala 2.13 version. */
+/** Trait that includes types used for cross-building compatibility; scala 2.12 version. */
 private[commons] trait Compat {
-  /** For annotations available at runtime in 2.13 we use ConstantAnnotation. */
+  /** For annotations available at runtime in 2.12 we use ClassfileAnnotation which has been deprecated in
+    * 2.13 and replaced with the newly defined ConstantAnnotation.. */
   type ConstantAnnotation = scala.annotation.ClassfileAnnotation
 
-  /** For basic views we use the new scala.collection.View. */
+  /** For basic views in 2.12 we use the old scala.collection.IterableView which has disappeared in 2.13 to
+    * be replaced by View. */
   type View[+A] = scala.collection.IterableView[A, Iterable[A]]
 }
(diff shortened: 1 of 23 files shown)


tfenne:
@nh13 I was playing around this morning with trying to eliminate the remaining deprecations caused by `ClassFileAnnotation` being deprecated.  The solution in 2.13 is to use `ConstantAnnotation` instead.  The problem is that that type does exist in 2.12.  For those I think we have two options:

1. Leave it as `ClassFileAnnotation` for now, knowing that in the future when we drop 2.12 support we can migrate over to `ConstantAnnotation`
2. Add a scala-2.12 and scala-2.13 source root and define a new type (perhaps in CommonsDef):

```
type RuntimeAnnotation = ClassfileAnnotation // in the 2.12 source root
type RuntimeAnnotation = ConstantAnnotation // in the 2.13 source root
```

Thoughts?

tfenne:
On further reflection, I think it's probably best to include split source roots down here in commons, and use those to handle everything we need for cross-building here and then also in sopt, and fgbio.  With the changes I've included here we can eliminate a lot of deprecation warnings, have source look more 2.13-esque _and_ remove some cross-building complexity in the PR I have open in fgbio.

src/main/scala-2.12-/com/fulcrumgenomics/commons/Compat.scala
```diff
@@ -0,0 +1,12 @@
...
+
+import scala.annotation.ClassfileAnnotation
+
+/** Trait that includes types used for cross-building compatibility; scala 2.13 version. */
+private[commons] trait Compat {
+  /** For annotations available at runtime in 2.13 we use ConstantAnnotation. */
+  type ConstantAnnotation = scala.annotation.ClassfileAnnotation
+
+  /** For basic views we use the new scala.collection.View. */
+  type View[+A] = scala.collection.IterableView[A, Iterable[A]]
```
tfenne, line 13: This isn't strictly needed for commons, but we need it in fgbio, so I think defining it here is a good idea.

src/main/scala/com/fulcrumgenomics/commons/collection/BetterBufferedIterator.scala
```diff
@@ -26,6 +26,8 @@ package com.fulcrumgenomics.commons.collection
 
 import com.fulcrumgenomics.commons.CommonsDef._
 
+import scala.collection.BufferedIterator
```
nh13, line 29: ?

nh13 (approved):
LGTM

src/main/scala/com/fulcrumgenomics/commons/collection/BetterBufferedIterator.scala
tfenne: I think scala.BufferedIterator got removed as part of the collections re-org, so now you have to import it specifically.

PR #57, opened 2019-08-18
Add support for reading from Unix/Posix pipes (fifos)

### LINKED ISSUE #56: Io.readLines should be able to read from a file descriptor
I'd like to support this: `fgbio Tool --input <(cat file1.gz file2.gz)`.  Unfortunately, we run into the same issue as this: https://github.com/samtools/htsjdk/issues/1084.  For `/dev/stdin` we have a [work around](https://github.com/fulcrumgenomics/commons/blob/master/src/main/scala/com/fulcrumgenomics/commons/io/Io.scala#L52), but if it is a file descriptor (ex. `/dev/fd/1234`) we hit the htsjdk issue above.

nh13:
Fixes #56 

Fgbio tools should now work with this too:
```
fgbio Tool --input <(cat fastq1.gz fastq2.gz)
```


src/main/scala/com/fulcrumgenomics/commons/io/Io.scala
```diff
@@ -49,8 +50,13 @@ trait IoUtil {
 
   /** Creates a new InputStream to read from the supplied path. */
   def toInputStream(path: Path) : InputStream = {
-    val stream =  if (Files.isSameFile(path, Io.StdIn)) System.in else Files.newInputStream(path)
-    new BufferedInputStream(stream, bufferSize)
+    if (Files.isSameFile(path, Io.StdIn)) System.in
+    else {
+      val stream = Files.newInputStream(path)
+      val attrs  = Files.readAttributes(path, classOf[BasicFileAttributes])
+      if (!attrs.isRegularFile && !attrs.isSymbolicLink && attrs.isOther) stream
```
tfenne: Can you add some comments here so we don't re-break this?  I'm curious why buffering those streams causes a problem, since it's also possible client code will attempt to wrap the input stream in a buffer.

src/test/scala/com/fulcrumgenomics/commons/io/IoTest.scala
```diff
@@ -62,6 +75,15 @@ class IoTest extends UnitSpec {
     Io.assertReadable(Io.StdIn)
   }
 
+  it should "not throw an exception for Unix fifos" in { if (hasMkfifo) {
+    val lines       = Seq("foo", "bar")
+    val pipePath    = Paths.get("test1").toAbsolutePath
```
tfenne: Can you use either `makeTempFile` or something else that will create files in a temp directory, not in the directory the test is running in please?

tfenne (changes requested):
@nh13 From reading the previous HTSJDK issue I'm not sure this is a great solution.  If the client wraps this in any kind of stream that uses `.available()` it'll still fail I believe.  I think that rather than not buffering the stream, we should change the code in `Io.toInputStream()` to fall back to `new FileInputStream(path.toFile)` if it detects that the input appears to be a non-regular file.  Thoughts?

nh13:
@tfenne back to you.

src/main/scala/com/fulcrumgenomics/commons/io/Io.scala
```diff
@@ -49,8 +50,28 @@ trait IoUtil {
...
+    //   at sun.nio.cs.StreamDecoder.implRead(StreamDecoder.java:326)
+    //   at sun.nio.cs.StreamDecoder.read(StreamDecoder.java:178)
+    //   at java.io.InputStreamReader.read(InputStreamReader.java:184)
+    //   at java.io.BufferedReader.fill(BufferedReader.java:161)
+    // This is a known issue; see: https://github.com/samtools/htsjdk/issues/1084
+    if (Files.isSameFile(path, Io.StdIn)) System.in
+    else {
+      val stream = Files.newInputStream(path)
+      val attrs  = Files.readAttributes(path, classOf[BasicFileAttributes])
+      if (!attrs.isRegularFile && !attrs.isSymbolicLink && attrs.isOther) stream
```
tfenne: Would it make sense to resolve the Path fully using `Path.toRealPath()` first?  And then we can just worry about what the target is and not whether we were handed a symlink?  I think specifically I would suggest changing line 71 to be `val attrs = Files.readAttributes(path.toRealPath(), ...` so that if there is any issue opening the path on line 70 that the user provided path is reported in the exception.

src/test/scala/com/fulcrumgenomics/commons/io/IoTest.scala
```diff
@@ -62,6 +75,15 @@ class IoTest extends UnitSpec {
     Io.assertReadable(Io.StdIn)
   }
 
+  it should "not throw an exception for Unix fifos" in { if (hasMkfifo) {
```
tfenne: Is there any way to do something like `(if (hasMakfifo) it else ignore) should ...` so that it's clear the test is ignored when `mkfifo` isn't present?

tfenne (approved):
A couple of suggestions.  Feel free to take or leave and then merge when you're done.

src/test/scala/com/fulcrumgenomics/commons/io/IoTest.scala
sstadick: I went with `assume(hasMkfifo)`

src/main/scala/com/fulcrumgenomics/commons/io/Io.scala
sstadick: I went with this

sstadick:
@nh13 @tfenne, I made the changes suggested by Tim, and merged in the recent other changes to the IO class. The tests pass and it looks like it works to me. Was there something deeper that needed doing for this? 

I'm not sure what triggered me becoming a reviewer on this, but it won't let me un-add myself :(

src/test/scala/com/fulcrumgenomics/commons/io/IoTest.scala
```diff
@@ -189,6 +212,19 @@ class IoTest extends UnitSpec {
...
+  "Io.readLines" should "read Unix fifos" in { 
+    assume(hasMkfifo, "Canceling Posix specific test.")
+    val lines    = Seq("foo", "bar")
+    val pipeDir  = tmpdir()
+    val pipePath = pipeDir.resolve("pipe2")
+    Files.deleteIfExists(pipePath) // in case of a failed previous test that needs clean up
+    pipe(pipePath, lines)
+    val writeFuture = Future { Io.writeLines(pipePath, lines); true }
+    Io.readLines(pipePath).toList should contain theSameElementsInOrderAs lines
+    Await.result(writeFuture, Duration(1, TimeUnit.SECONDS)) shouldBe true
```
tfenne: I'd maybe bump up the wait time here to a few seconds just so we don't get sporadic test failures if the travis instance is slow or pauses.

tfenne (approved):
I think this looks good to me.  I'm struggling to remember all the context, but if the tests are passing I think we're in fairly good shape.  The only other suggestion I would have is to perhaps test that `toInputStream` returns `BufferedInputStream`s when called on both regular files and symlinks to regular files, just to be sure we don't generate a performance regression.

sstadick:
@tfenne @nh13, related to the support of gzipped files, what is your take on using the magic bytes at the start of the input to determine the encoding? That would allow this to not be based on file extensions and work for detecting the encoding type of piped in streams as well? 

Example I've seen in the wild (and only one): https://github.com/aseyboldt/fastq-rs/blob/5f0d3ab0515f060dfd1c16f65d52ed80d7043574/src/lib.rs#L194

List of magic bytes: https://gist.github.com/leommoore/f9e57ba2aa4bf197ebc5

I suppose this would be a separate issue anyways.

src/test/scala/com/fulcrumgenomics/commons/io/IoTest.scala
sstadick: Bumped up up time of waiting, added test for BufferedInputStream for both symlink and real files

nh13:
@sstadick I think it is a good idea, but perhaps from another PR. When in doubt, check htsjdk: https://github.com/samtools/htsjdk/blob/e357c42d6e31eb4335fd53496d097f0186276dd7/src/main/java/htsjdk/samtools/util/IOUtil.java#L1310-L1329

I think the major issue is we do not want htsjdk as a dependency in commons.

nh13:
@sstadick before you merge this, could you build a version of fgbio with this version and test a tool?  https://github.com/fulcrumgenomics/commons/issues/56

sstadick:
@nh13 Tested in real life via SortBam and it successfully read from the pipe

```
java -jar ./target/scala-2.13/fgbio-1.2.0-afc95c8-SNAPSHOT.jar SortBam -i <(cat src/test/resources/com/fulcrumgenomics/bam/set_mate_coordinatesorted.sam ) -o temp.sam
```

src/main/scala/com/fulcrumgenomics/commons/io/Io.scala
```diff
@@ -55,12 +56,16 @@ trait IoUtil {
...
   /** Creates a new InputStream to read from the supplied path. Automatically handles gzipped files. */
   def toInputStream(path: Path) : InputStream = {
+
     PathUtil.extensionOf(path) match {
       case Some(".gz") | Some(".bgz") | Some(".bgzip") =>
         new GZIPInputStream(Files.newInputStream(path), bufferSize)
       case _ => {
         val stream =  if (Files.isSameFile(path, Io.StdIn)) System.in else Files.newInputStream(path)
+        val attrs  = Files.readAttributes(path.toRealPath(), classOf[BasicFileAttributes])
         new BufferedInputStream(stream, bufferSize)
```
tfenne: Shouldn't this line go?  It's constructing then throwing away a buffered input stream?

src/main/scala/com/fulcrumgenomics/commons/io/Io.scala
```diff
@@ -55,12 +56,16 @@ trait IoUtil {
...
+        if (!attrs.isRegularFile && attrs.isOther) stream
```
tfenne: Can you add a comment here that this is to support named pipes?  And I have no recollection why we don't buffer named pipes, do you?

src/test/scala/com/fulcrumgenomics/commons/io/IoTest.scala
```diff
@@ -53,9 +59,13 @@ class IoTest extends UnitSpec {
     path
   }
 
-  /** Impl of IoUtil to test that compressionLevel can be overridden and set */
-  class FakeIo(var compressionLevel: Int = 5, override val bufferSize: Int = 128*1024) extends IoUtil {}
-  object FakeIo extends FakeIo(compressionLevel=5, bufferSize=128*1024)
+  private val hasMkfifo: Boolean = FileSystems.getDefault().supportedFileAttributeViews().contains("posix")
+
+  def pipe(path: Path, lines: Seq[String]): Unit = {
```
tfenne: lines is never used here

tfenne (approved):
Once comments addressed

PR #58, opened 2019-09-06
Add a general CommandlineTool

fangylo:
## Background

This PR was motivated by the need to use command line executables and have the ability to:
- Create a shell command task wrapping this executable
- Only run the task if this executable is available on the system path

`Fgbio` already has the tool [`Rscript`](https://github.com/fulcrumgenomics/fgbio/blob/master/src/main/scala/com/fulcrumgenomics/util/Rscript.scala) that achieves the same functionality for `R`. To reduce code duplication here we'd like to make a general `CommandLineTool` with the ability to extend to different executables such as R, python, Ghostscript, etc.

@nh13 suggested instead of putting this general `CommandLineTool` into `fgbio`, we should put it into `commons` and deprecate the original [Rscript]((https://github.com/fulcrumgenomics/fgbio/blob/master/src/main/scala/com/fulcrumgenomics/util/Rscript.scala)) in `fgbio`. 

## Changes made in this PR

1. Created a `CommandLineTool` that has some basic functionality to run a command line executable and the possibility to be extended. Current functions include:
	- Check if the executable is available
	- Check if certain modules is installed for the executable, e.g. R or python modules.
	- Run the executable if it is available.

2. Addressed the issue raised by @tfenne and @nh13 [here](https://github.com/fulcrumgenomics/fgbio/issues/397) and had `exec` and `execIfAvailable` both return `Unit` instead of `Try[Unit]`.

3. Currently [`Rscript.Available`](https://github.com/fulcrumgenomics/fgbio/blob/f05508919b647cc086dfc954e6192365e6d58b36/src/main/scala/com/fulcrumgenomics/util/Rscript.scala#L46) only returns **TRUE** when *both Rscript is available AND ggplot2 is installed*. This is a bit confusing so a separate `Rscript.ggplot2Available` is created to check the above instead. We can later deprecate the original `Rscript.Available`.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/commons/io/CommandLineTool.scala
+++ b/src/main/scala/com/fulcrumgenomics/commons/io/CommandLineTool.scala
@@ -24,68 +24,103 @@
 
 package com.fulcrumgenomics.commons.io
 
-import java.nio.file.{Files, Path}
+import java.nio.file.Path
 
 import com.fulcrumgenomics.commons.util.LazyLogging
 
-import scala.collection.mutable
-import scala.io.Source
-import scala.util.{Failure, Success, Try}
+import scala.collection.mutable.ListBuffer
+import scala.util.Try
 
 trait CommandLineTool extends LazyLogging {
+
   /** The name of the executable such as Rscript or gs. */
   val Executable: String
 
-  /** Command used for the above executable. */
-  val TestCommand: Seq[String]
+  /** Command used for the above executable.
+    *
+    * For example, to run `Rscript --version`:
+    * {{{
+    * scala> import com.fulcrumgenomics.commons.io._
+    * scala> val TestArgs = Seq("--version")
+    * scala> val Executable = "Rscript"
+    * scala> Rscript.execArgs(Executable +: TestArgs:_*)
+    * }}}
+    * */
+  val TestArgs: Seq[String]
 
   /** Exception class that holds onto an executable's exit/status code. */
   case class ToolException(status: Int) extends RuntimeException {
     override def getMessage: String = s"$Executable failed with exit code $status."
   }
 
-  /** Returns a tuple of (Boolean, String): first element is true when command can execute
-    * with no error; second element is a string of command stdout */
-  def execArgs(command: String*): (Boolean, String) = {
-    try {
+  /** Executes a command and returns the stdout.
+    *
+    * @param command the command to be executed
+    */
+  def execArgs(command: String*): Try[ListBuffer[String]] = {
+    Try {
       // Read redirected stdout to a string
-      val process = new ProcessBuilder(command: _*).redirectErrorStream(true).start()
-      val stdout  = mutable.ListBuffer[String]()
-      val sink    = new AsyncStreamSink(process.getInputStream, s => stdout.append(s))
-      // True if the command can execute with no error
-      val canExecute = process.waitFor() == 0
+      val process  = new ProcessBuilder(command: _*).redirectErrorStream(true).start()
+      val stdout   = ListBuffer[String]()
+      val sink     = new AsyncStreamSink(process.getInputStream, s => stdout.append(s))
+      val exitCode = process.waitFor()
+
       sink.close()
 
-      (canExecute, stdout.toList.mkString)
+      require(exitCode == 0, s"Command did not execute successfully. Exit code: $exitCode")
+
+      stdout
     }
-    catch {case e: Exception => (false, "")}
   }
 
   /** Returns true if the tool is available and false otherwise. */
-  lazy val Available: Boolean = {
-    execArgs(Executable +: TestCommand: _*)._1
-  }
+  lazy val Available: Boolean = execArgs(Executable +: TestArgs: _*).isSuccess
 }
 
-trait CanRunScript {
+/** Trait that indicates the executable can run scripts.*/
+trait ScriptRunner {
   self: CommandLineTool =>
+
   /** Suffix for scripts that can be run by this tool. */
   val Suffix: String
 
-  /** Executes a script from the classpath if the tested executable is available. */
-  def execIfAvailable(scriptResource: String, args: String*): Try[Unit] =
-    if (Available) exec(scriptResource, args:_*) else Success(Unit)
+  /** Executes a script from the classpath if this executable is available.
+    *
+    * @param scriptResource the name of the script resource on the classpath
+    * @param args a variable list of arguments to pass to the script
+    * @throws Exception when we are unable to execute to this script on the classpath with the given arguments.
+    */
+  def execIfAvailable(scriptResource: String, args: String*): Unit =
+    try { if (Available) exec(scriptResource, args:_*) }
+    catch { case e: Exception => throw new Exception(s"Cannot execute script from $scriptResource with args:$args") }
 
-  /** Executes from a script stored at a Path if the tested executable is available. */
-  def execIfAvailable(script: Path, args: String*): Try[Unit] =
-    if (Available) exec(script, args:_*) else Success(Unit)
+  /** Executes a script stored at a Path if the this executable is available.
+    *
+    * @param script Path of the script to be run
+    * @param args a variable list of arguments to pass to the script
+    * @throws Exception when we are unable to execute to this script on the classpath with the given arguments.
+    */
+  def execIfAvailable(script: Path, args: String*): Unit = execIfAvailable(script.toString, args: _*)
 
-  /** Executes a script from the classpath. */
-  def exec(scriptResource: String, args: String*): Try[Unit] =
-    Try { writeResourceToTempFile(scriptResource) }.map(path => exec(path, args:_*))
+  /** Executes a script from the classpath if this executable is available, raise an exception otherwise.
+    *
+    * @throws ToolException when the exit code from the called process is not zero.
+    * @throws Exception when we are unable to execute to this script on the classpath with the given arguments.
+    * @param scriptResource the name of the script resource on the classpath
+    * @param args a variable list of arguments to pass to the script
+    * @return Unit as this code is side-effecting
+    */
+  def exec(scriptResource: String, args: String*): Unit = exec(writeResourceToTempFile(scriptResource), args:_*)
 
-  /** Executes from a script stored at a Path. */
-  def exec(script: Path, args: String*): Try[Unit] = Try {
+  /** Execute from a script stored at a Path if this executable is available, raise an exception otherwise.
+    *
+    * @throws ToolException when the exit code from the called process is not zero.
+    * @throws Exception when we are unable to execute to this script with the given arguments.
+    * @param script Path of the script to be run
+    * @param args a variable list of arguments to pass to the script
+    * @return Unit as this code is side-effecting
+    */
+  def exec(script: Path, args: String*): Unit = try {
     val command = Executable +: script.toAbsolutePath.toString +: args
     val process = new ProcessBuilder(command:_*).redirectErrorStream(false).start()
     val pipe1   = Io.pipeStream(process.getErrorStream, logger.info)
@@ -95,9 +130,13 @@ trait CanRunScript {
     pipe2.close()
 
     if (retval != 0) throw ToolException(retval)
-  }
+  } catch {case e: Exception => throw new Exception(s"Cannot execute script from path $script with args: $args") }
 
-  /** Extracts a resource from the classpath and writes it to a temp file on disk. */
+  /** Extract a resource from the classpath and writes it to a temp file on disk.
+    *
+    * @param resource a given name on the classpath
+    * @return path to the temporary file
+    * */
   private def writeResourceToTempFile(resource: String): Path = {
     val lines = Io.readLinesFromResource(resource).toSeq
     val path  = Io.makeTempFile("script.", suffix = Suffix)
@@ -109,13 +148,16 @@ trait CanRunScript {
 
 trait Versioned {
   self: CommandLineTool =>
-  val VersionFlag: String      = "--version"
-  val TestCommand: Seq[String] = Seq(VersionFlag)
 
-  /** Returns version of the tool */
-  lazy val Version: String = execArgs(Executable +: TestCommand:_*)._2
+  /** The default version flag. */
+  val VersionFlag: String   = "--version"
+
+  /** Argument used to test version of the executable. */
+  val TestArgs: Seq[String] = Seq(VersionFlag)
+
+  /** Return version of this tool */
+  lazy val Version: Try[ListBuffer[String]] = execArgs(Executable +: TestArgs:_*)
 }
-
 
 trait Modular {
   self: CommandLineTool =>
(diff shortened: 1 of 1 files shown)


fangylo:
Looks like integration test failed because of `par`:
https://github.com/fulcrumgenomics/commons/blob/521c91ee15da0e054bd227d1195a3dffa50f2195/src/main/scala/com/fulcrumgenomics/commons/io/CommandLineTool.scala#L184

In scala 2.13 we need to add 
```
import scala.collection.parallel.CollectionConverters._
```
but that only works for 2.13 not 2.12...shall we avoid using `par` here? 
https://github.com/scala/scala-parallel-collections#cross-building-source-compatibility

src/main/scala/com/fulcrumgenomics/commons/io/CommandLineTool.scala
```diff
@@ -0,0 +1,216 @@
...
+    val command = Executable +: script.toAbsolutePath.toString +: args
+    val process = new ProcessBuilder(command:_*).redirectErrorStream(false).start()
+    val pipe1   = Io.pipeStream(process.getErrorStream, logger.info)
+    val pipe2   = Io.pipeStream(process.getInputStream, logger.debug)
+    val retval  = process.waitFor()
+    pipe1.close()
+    pipe2.close()
+
+    if (retval != 0) throw ToolException(retval)
+  } catch {case e: Exception => throw new Exception(s"Cannot execute script from path $script with args: $args") }
```
clintval: The prior line (131) will throw a specific exception if the return value is non-zero. However, on this line we catch all exceptions and throw away our custom `ToolException`.

Exception handling could use some work here.

src/main/scala/com/fulcrumgenomics/commons/io/CommandLineTool.scala
```diff
@@ -0,0 +1,216 @@
...
+    *
+    * @throws ToolException when the exit code from the called process is not zero
+    * @throws Exception when we are unable to execute the script with the given arguments
+    * @param script Path to the script to be executed
+    * @param args a variable list of arguments to pass to the script
+    */
+  def exec(script: Path, args: String*): Unit = try {
+    val command = Executable +: script.toAbsolutePath.toString +: args
+    val process = new ProcessBuilder(command:_*).redirectErrorStream(false).start()
+    val pipe1   = Io.pipeStream(process.getErrorStream, logger.info)
```
clintval: I think this should be:

```suggestion
    val pipe1   = Io.pipeStream(process.getErrorStream, logger.Warning)
```

src/main/scala/com/fulcrumgenomics/commons/io/CommandLineTool.scala
```diff
@@ -0,0 +1,216 @@
...
+    val pipe2   = Io.pipeStream(process.getInputStream, logger.debug)
```
clintval: I think this should be:

```suggestion
    val pipe2   = Io.pipeStream(process.getInputStream, logger.info)
```

clintval: Also, do we want `getOutputStream` here?

clintval (commented):
Few comments after using the `Python3` object to execute scripts from Scala.

src/main/scala/com/fulcrumgenomics/commons/io/CommandLineTool.scala
```diff
@@ -0,0 +1,216 @@
...
+  case class ToolException(status: Int) extends RuntimeException {
+    override def getMessage: String = s"$Executable failed with exit code $status."
+  }
+
+  /** Executes a command and returns the stdout.
+    *
+    * @param command the command to be executed
+    */
+  def execCommand(command: String*): Try[ListBuffer[String]] = {
+    Try {
```
clintval, line 71: I would appreciate a `logging.debug("Executing: " + command.mkString(" "))` here.

src/main/scala/com/fulcrumgenomics/commons/io/CommandLineTool.scala
```diff
@@ -0,0 +1,216 @@
...
+
+  /** Executes a script from the filesystem Path.
+    *
+    * @throws ToolException when the exit code from the called process is not zero
+    * @throws Exception when we are unable to execute the script with the given arguments
+    * @param script Path to the script to be executed
+    * @param args a variable list of arguments to pass to the script
+    */
+  def exec(script: Path, args: String*): Unit = try {
+    val command = Executable +: script.toAbsolutePath.toString +: args
```
clintval: Similarily, would appreciate this here:

```scala
logging.debug("Executing $script with $Executable: args.mkString(" ")")
```

src/main/scala/com/fulcrumgenomics/commons/io/CommandLineTool.scala
```diff
@@ -0,0 +1,216 @@
...
+
+    if (retval != 0) throw ToolException(retval)
+  } catch {case e: Exception => throw new Exception(s"Cannot execute script from path $script with args: $args") }
+
+  /** Extracts a resource from the classpath and writes it to a temp file on disk.
+    *
+    * @param resource a given name on the classpath
+    * @return path to the temporary file
+    * */
+  private def writeResourceToTempFile(resource: String): Path = {
```
clintval, line 146: To make logging of script execution more effective, I use this function instead, which utilizes the filename of the original script resource:

```scala
  private def writeResourceToTempFile(resource: String): Path = {
    val lines = Io.readLinesFromResource(resource)
    val dir   = Io.makeTempDir(PathUtil.sanitizeFileName(this.getClass.getSimpleName))
    val path  = PathUtil.pathTo(dir.toString, PathUtil.basename(resource, trimExt = false))
    Seq(dir, path).foreach(_.toFile.deleteOnExit())
    Io.writeLines(path, lines)
    path
  }
```

src/main/scala/com/fulcrumgenomics/commons/io/CommandLineTool.scala
```diff
@@ -0,0 +1,216 @@
...
+
+import java.nio.file.Path
+
+import com.fulcrumgenomics.commons.util.LazyLogging
+import com.fulcrumgenomics.commons.CommonsDef._
+
+import scala.collection.mutable.ListBuffer
+import scala.util.Try
+
+/** Provides methods for an executable to run it. For example, methods to run this
```
nh13:
```suggestion
/** Provides methods to run an executable. For example, methods to run this
```

src/main/scala/com/fulcrumgenomics/commons/io/CommandLineTool.scala
```diff
@@ -0,0 +1,216 @@
...
+  * executable using provided arguments, test if this executable is available, etc.
+  */
+trait CommandLineTool extends LazyLogging {
+
+  /** The name of the executable. */
+  val Executable: String
```
nh13:
```suggestion
  val executable: String
```
variables should be lower camel case, except for constants (which are part of objects): https://docs.scala-lang.org/style/naming-conventions.html

src/main/scala/com/fulcrumgenomics/commons/io/CommandLineTool.scala
```diff
@@ -0,0 +1,216 @@
...
+
+  /** Arguments for the above executable.
+    *
+    * For example, to run `Rscript --version`:
+    * {{{
+    * scala> import com.fulcrumgenomics.commons.io._
+    * scala> val TestArgs = Seq("--version")
+    * scala> val Executable = "Rscript"
```
nh13: lower camel case here and above

src/main/scala/com/fulcrumgenomics/commons/io/CommandLineTool.scala
```diff
@@ -0,0 +1,216 @@
...
+    * }}}
+    * */
+  val TestArgs: Seq[String]
+
+  /** Exception class that holds onto the exit/status code of the executable. */
+  case class ToolException(status: Int) extends RuntimeException {
+    override def getMessage: String = s"$Executable failed with exit code $status."
+  }
+
+  /** Executes a command and returns the stdout.
```
nh13: stderr is redirected no?

src/main/scala/com/fulcrumgenomics/commons/io/CommandLineTool.scala
```diff
@@ -0,0 +1,216 @@
...
+      val sink     = new AsyncStreamSink(process.getInputStream, s => stdout.append(s))
+      val exitCode = process.waitFor()
+
+      require(exitCode == 0, s"Command did not execute successfully. Exit code: $exitCode")
+      yieldAndThen(stdout)(sink.safelyClose())
+    }
+  }
+
+  /** Returns true if the tool is available and false otherwise. */
+  lazy val Available: Boolean = execCommand(Executable +: TestArgs: _*).isSuccess
```
nh13:
```suggestion
  lazy val available: Boolean = execCommand(Executable +: TestArgs: _*).isSuccess
```

src/main/scala/com/fulcrumgenomics/commons/io/CommandLineTool.scala
```diff
@@ -0,0 +1,216 @@
...
+      val stdout   = ListBuffer[String]()
+      val sink     = new AsyncStreamSink(process.getInputStream, s => stdout.append(s))
+      val exitCode = process.waitFor()
+
+      require(exitCode == 0, s"Command did not execute successfully. Exit code: $exitCode")
+      yieldAndThen(stdout)(sink.safelyClose())
+    }
+  }
+
+  /** Returns true if the tool is available and false otherwise. */
```
nh13:
```suggestion
  /** True if the tool is available and false otherwise. */
```
It's a `val`, so it's either true or false, and returns nothing

src/main/scala/com/fulcrumgenomics/commons/io/CommandLineTool.scala
```diff
@@ -0,0 +1,216 @@
...
+  lazy val Available: Boolean = execCommand(Executable +: TestArgs: _*).isSuccess
+}
+
+/** Indicates the executable can run scripts.*/
+trait ScriptRunner {
+  self: CommandLineTool =>
+
+  /** Suffix of scripts that can be run by this tool. */
+  val Suffix: FilenameSuffix
```
nh13:
```suggestion
  val suffix: FileNameSuffix
```
I'm going to stop commenting each place  the `val` should be lower camel case.  I'll re-check after the next round of changes.

src/main/scala/com/fulcrumgenomics/commons/io/CommandLineTool.scala
```diff
@@ -0,0 +1,216 @@
...
+  /** Executes a script from the classpath, raise an exception otherwise.
+    *
+    * @throws ToolException when the exit code from the called process is not zero.
+    * @throws Exception when we are unable to execute to this script on the classpath with the given arguments.
+    * @param scriptResource the name of the script resource on the classpath
+    * @param args a variable list of arguments to pass to the script
+    */
+  def exec(scriptResource: String, args: String*): Unit = exec(writeResourceToTempFile(scriptResource), args: _*)
+
+  /** Executes a script from the filesystem Path.
```
nh13:
```suggestion
  /** Executes a script from the filesystem path.
```

src/main/scala/com/fulcrumgenomics/commons/io/CommandLineTool.scala
nh13: Also, this needs to be split across multiple lines

src/main/scala/com/fulcrumgenomics/commons/io/CommandLineTool.scala
```diff
@@ -0,0 +1,216 @@
...
+trait Versioned {
+  self: CommandLineTool =>
+
+  /** The default version flag. */
+  val VersionFlag: String   = "--version"
+
+  /** Argument used to test version of the executable. */
+  val TestArgs: Seq[String] = Seq(VersionFlag)
+
+  /** Returns version of this executable. */
```
nh13: Ditto about returns, fix everywhere

src/main/scala/com/fulcrumgenomics/commons/io/CommandLineTool.scala
```diff
@@ -0,0 +1,216 @@
...
+
+/** Defines tools to test various version of python executables */
+trait Python extends CommandLineTool with Versioned with Modular with ScriptRunner {
+  val Executable: String
+  val Suffix: FilenameSuffix = ".py"
+  def TestModuleCommand(module: String): Seq[String] = Seq(Executable, "-c", s"import $module")
+}
+
+/** Objects specific for different Python versions.  */
+object Python extends Python  { val Executable: String = "python" }
```
nh13: how about an apply method instead?
```scala
object Python {
  def apply(version: String = ""): Python = new Python { val Executable: String = f"python$version" }
  val Python2 = Python("2")
  val Python3 = Python("3")
}
```

nh13 (changes requested):
Minor quibbles about variable naming.  I also agree with @clintval's comments.  This also needs some tests!

nh13:
@fangylo I forgot about your `.par` question.  I'd use the various `Par*` classes and methods in https://github.com/fulcrumgenomics/commons/blob/master/src/main/scala/com/fulcrumgenomics/commons/CommonsDef.scala based on what you're trying to do.  Is there a good reason to use `.par` here, as the parallelism doesn't look configurable.

src/main/scala/com/fulcrumgenomics/commons/io/CommandLineTool.scala
fangylo: @nh13 to achieve this the original `Python` needs to be changed from `trait` to a `class`. Since it probably doesn't make a lot of sense to have multiple `Python`s, I made the class constructor _private_ so that only its companion object can instantiate instances. I'm not sure if this is better or the original is simpler/ clearer though.

fangylo: Been going back and forth with this. Originally I changed to this block of code according to @nh13 's suggestion:
```
/** Defines tools to test various version of python executables */
class Python (val pythonVersion: String, val executable: String = "python") extends
  CommandLineTool with Versioned with Modular with ScriptRunner {
  val suffix: FilenameSuffix = ".py"
  def TestModuleCommand(module: String): Seq[String] = Seq(executable, "-c", s"import $module")
}

/** Objects specific for different Python versions.  */
object Python {
  def apply(pythonVersion : String = "", executable: String = "python" ) =
    new Python(pythonVersion, f"$executable$pythonVersion")
  val Python2 = new Python("2")
  val Python3 = new Python("3")
}
```
However, there is concern that we probably want `Python` to be singleton. Therefore this is changed back to `trait` at the moment. Major `Python` versions are traits extended from `Python`.

fangylo:
@nh13 Thanks for the link for `par`. The reason to use `par` here is that we'd like to be able to check the availability of different modules in parallel. This is currently default to all available cores but can be configured.

src/main/scala/com/fulcrumgenomics/commons/io/CommandLineTool.scala
```diff
@@ -0,0 +1,231 @@
...
+
+  /** Command to test if a R module exists.
+    *
+    * @param module name of the module to be tested
+    * @return command used to test if a given R module is installed
+    * */
+  def TestModuleCommand(module: String): Seq[String] = Seq(executable, "-e", s"stopifnot(require('$module'))")
+
+  /** True if Rscript exists. */
+  override lazy val available: Boolean = execCommand(executable, versionFlag).isSuccess
```
clintval, line 210: I think you can safely delete this method as the overridden behavior is the same as the underlying behavior.

src/main/scala/com/fulcrumgenomics/commons/io/CommandLineTool.scala
```diff
@@ -0,0 +1,231 @@
...
+ * THE SOFTWARE.
+ */
+package com.fulcrumgenomics.commons.io
+
+import java.nio.file.Path
+
+import com.fulcrumgenomics.commons.CommonsDef
+import com.fulcrumgenomics.commons.util.LazyLogging
+import com.fulcrumgenomics.commons.CommonsDef._
+//import com.fulcrumgenomics.commons.io.Python.executable
```
clintval, line 31: del

src/main/scala/com/fulcrumgenomics/commons/io/CommandLineTool.scala
```diff
@@ -0,0 +1,231 @@
...
+  /** Version of this executable. */
+  lazy val version: Try[ListBuffer[String]] = execCommand(executable +: testArgs: _*)
+}
+
+/** Defines methods used to check if specific modules are installed with the executable . */
+trait Modular {
+  self: CommandLineTool =>
+
+  /** Command used to test if a module is included with the tested executable. */
+  def TestModuleCommand(module: String): Seq[String]
```
clintval, line 175:
```suggestion
  def testModuleCommand(module: String): Seq[String]
```
https://docs.scala-lang.org/style/naming-conventions.html#constants-values-variable-and-methods

Likewise, lowercase:

- `ModuleAvailable`

src/main/scala/com/fulcrumgenomics/commons/io/CommandLineTool.scala
```diff
@@ -0,0 +1,231 @@
...
+
+/** Major Python versions as traits.  */
+trait Python2 extends Python { override val executable = "python2"}
+trait Python3 extends Python { override val executable = "python3"}
+
+/** Popular major and minor Python versions.  */
+object Python2 extends Python2
+object Python3 extends Python3
+object Python2_7 extends Python2  { override val executable: String = "python2.7" }
+object Python3_7 extends Python3  { override val executable: String = "python3.7" }
```
clintval, line 231: Python3.6 is probably more popular than 3.7 at this point in time, could include both?

src/main/scala/com/fulcrumgenomics/commons/io/CommandLineTool.scala
```diff
@@ -0,0 +1,231 @@
...
+  /** Exception class that holds onto the exit/status code of the executable. */
+  case class ToolException(status: Int) extends RuntimeException {
+    override def getMessage: String = s"$executable failed with exit code $status."
+  }
+
+  /** Executes a command and returns the stdout and stderr.
+    *
+    * @param command the command to be executed
+    */
+  def execCommand(command: String*): Try[ListBuffer[String]] = {
```
clintval, line 69: I've found keeping the streams separate more useful for shelling out and collecting output. I've since modified to:

```scala
case class ProcessOutput(
  stdout: ListBuffer[String] = ListBuffer[String](), // Is this a bad idiom?
  stderr: ListBuffer[String] = ListBuffer[String]()
)

/** Executes a command and returns the stdout and stderr.
  *
  * @param command the command to be executed
  */
def execCommand(command: String*): Try[ProcessOutput] = {
  logger.debug("Executing: " + command.mkString(" "))
  Try {
    val output   = ProcessOutput()
    val process  = new ProcessBuilder(command: _*).redirectErrorStream(false).start()
    val outSink  = new AsyncStreamSink(process.getInputStream, s => output.stdout.append(s))
    val errSink  = new AsyncStreamSink(process.getErrorStream, s => output.stderr.append(s))
    val exitCode = process.waitFor()
    if (exitCode != 0) throw ToolException(exitCode)
    Seq(outSink, errSink).foreach(_.safelyClose())
    output
  }
}
```

PR #59, opened 2019-10-08
Add SystemUtil for querying common system properties

clintval:
Functionality pulled from [`fgbio`](https://github.com/fulcrumgenomics/fgbio/blob/8f9d01ab4bc4b61e8e7ac696c5e2a91e3165f705/src/main/scala/com/fulcrumgenomics/cmdline/FgBioMain.scala#L46) and [`dagr`](https://github.com/fulcrumgenomics/dagr/pull/368).

Which of these would you like as a part of public API?

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/commons/util/SystemUtil.scala
+++ b/src/main/scala/com/fulcrumgenomics/commons/util/SystemUtil.scala
@@ -61,13 +61,13 @@ object SystemUtil {
   def isOsArch(name: String): Boolean = OsArch.contains(name)
 
   /** Returns true if the architecture matches the given regular expression, false otherwise. */
-  def isOsArch(regex: Regex): Boolean = OsArch.exists(regex.matches)
+  def isOsArch(regex: Regex): Boolean = OsArch.exists(v => regex.findFirstMatchIn(v).isDefined)
 
   /** Returns true if the operating system version is the given name, false otherwise. */
   def isOsVersion(name: String): Boolean = OsVersion.contains(name)
 
   /** Returns true if the operating system version matches the given regular expression, false otherwise. */
-  def isOsVersion(regex: Regex): Boolean = OsVersion.exists(regex.matches)
+  def isOsVersion(regex: Regex): Boolean = OsVersion.exists(v => regex.findFirstMatchIn(v).isDefined)
 
   /** True if the current system could support the Intel Inflater and Deflater, false otherwise. */
   lazy val IntelCompressionLibrarySupported: Boolean = {
--- a/src/main/scala/com/fulcrumgenomics/commons/util/SystemUtil.scala
+++ b/src/main/scala/com/fulcrumgenomics/commons/util/SystemUtil.scala
@@ -73,7 +73,7 @@ object SystemUtil {
   lazy val IntelCompressionLibrarySupported: Boolean = {
     if (!SystemUtil.IsOsLinux && !SystemUtil.IsOsMac) false
     else if (SystemUtil.isOsArch(name = "ppc64le")) false
-    else if (SystemUtil.IsOsMac && SystemUtil.isOsVersion("10\\.14\\.".r)) false // FIXME: https://github.com/Intel-HLS/GKL/issues/101
+    else if (SystemUtil.IsOsMac && SystemUtil.isOsVersion("10\\.14\\..+".r)) false // FIXME: https://github.com/Intel-HLS/GKL/issues/101
     else true
   }
 }
--- a/src/main/scala/com/fulcrumgenomics/commons/util/SystemUtil.scala
+++ b/src/main/scala/com/fulcrumgenomics/commons/util/SystemUtil.scala
@@ -1,7 +1,7 @@
 /*
  * The MIT License
  *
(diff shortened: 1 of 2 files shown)


nh13:
@clintval what do you think about these changes? https://github.com/clintval/commons/pull/1.  I'm happy to update fgbio and dagr for you.

clintval:
A+ go for it.

Was going to write tests after the public API was revealed, but thanks for beating me to it!

src/main/scala/com/fulcrumgenomics/commons/util/SystemUtil.scala
```diff
@@ -0,0 +1,79 @@
...
+  def isOsArch(name: String): Boolean = OsArch.contains(name)
+
+  /** Returns true if the architecture matches the given regular expression, false otherwise. */
+  def isOsArch(regex: Regex): Boolean = OsArch.exists(v => regex.findFirstMatchIn(v).isDefined)
+
+  /** Returns true if the operating system version is the given name, false otherwise. */
+  def isOsVersion(name: String): Boolean = OsVersion.contains(name)
+
+  /** Returns true if the operating system version matches the given regular expression, false otherwise. */
+  def isOsVersion(regex: Regex): Boolean = OsVersion.exists(v => regex.findFirstMatchIn(v).isDefined)
```
clintval: The member `Regex.matches()` does not exist in Scala 2.12, so I switched to `findFirstMatchIn()`.

But, that is not correct because we want full match anchoring!

For example. This is now possible, but shouldn't be:

```scala
scala> OsVersion
Some(0...0.00.0..0.10.14.6)
scala> isOsVersion("10\\.14\\.".r)
true
```

I'll try again.

clintval:
@nh13, all set. I found a way to have anchored regex matching that is 2.12 and 2.13 compatible. Ping me if you want help with the rest.

src/main/scala/com/fulcrumgenomics/commons/util/SystemUtil.scala
```diff
@@ -0,0 +1,79 @@
...
+  def getSystemProperty(property: String): Option[String] = {
+    try { Option(System.getProperty(property)) }
+    catch { case _: SecurityException => None } // not allowed to look at this property
+  }
+
+  /** True if this operating system is Linux, false otherwise. */
+  lazy val IsOsLinux: Boolean = LinuxNamePrefixes.exists(prefix => OsName.exists(_.startsWith(prefix)))
+
+  /** True if this operating system is Mac, false otherwise. */
+  lazy val IsOsMac: Boolean = MacNamePrefixes.exists(prefix => OsName.exists(_.startsWith(prefix)))
```
clintval: Would you rather us just use the Scala builtins?

https://www.scala-lang.org/api/current/scala/util/Properties$.html#isMac:Boolean

clintval (commented):
I would be interested in seeing this merged as I want to factor this into `PicardTask` in Dagr.

src/main/scala/com/fulcrumgenomics/commons/util/SystemUtil.scala
```diff
@@ -0,0 +1,79 @@
...
+
+  /** The current operating system architecture */
+  lazy val OsArch: Option[String] = getSystemProperty(property = "os.arch")
+
+  /** The current operating system version */
+  lazy val OsVersion: Option[String] = getSystemProperty(property = "os.version")
+
+  /** Gets a system property.  Returns None if not found or we are not allowed to retrieve the property. */
+  def getSystemProperty(property: String): Option[String] = {
+    try { Option(System.getProperty(property)) }
```
clintval: Could use builtin? https://www.scala-lang.org/api/current/scala/util/Properties$.html#isMac:Boolean

src/main/scala/com/fulcrumgenomics/commons/util/SystemUtil.scala
nh13: That makes a lot more sense, nice pickup

nh13 (changes requested):
One suggested change you yourself suggested.

clintval:
Few changes:

- Copyrights to 2020
- Changed `getSystemProperty` to `safePropOrNone` to match the ecosystem of Scala property getters in [`scala.util.Properties`](https://www.scala-lang.org/api/current/scala/util/Properties$.html#propIsSet(name:String):Boolean) for when you want to get a system property safely (catch security manager exceptions)
- Deleted `lazy val OsName` as that's in the [Scala builtins](https://www.scala-lang.org/api/current/scala/util/Properties$.html#osName:String)
- Moved over implementations to other Scala builtins like `isLinux`, `isMac`, and `propOrNone`

clintval:
@nh13 I think this is merge-ready FYI

clintval:
If possible, I'd like to find a way to move this along because I also want to use these features for patches to `fgbio` and `dagr`. I'm still having difficulty with the occasional core dump on MacOS (`10.15.3`) due to the Intel infalters/deflaters and want to unblock https://github.com/fulcrumgenomics/dagr/pull/368.

nh13 (approved):

PR #60, opened 2019-10-24
A lazy source.

nh13:


src/main/scala/com/fulcrumgenomics/commons/io/Io.scala
```diff
@@ -226,4 +226,10 @@ trait IoUtil {
     stream.safelyClose()
     array
   }
+
+  /** An iterable view of source data.  The transformations are non-strict, and the source is able to be closed.  */
+  trait LazySource[A] extends View[A] with Closeable {
```
nh13: go crazy with the name, it just can't be `Source`

tfenne: Hehehe.  I think the name is fine.  I think I would just extract it to a top level class and put it alongside `Writer`.

src/main/scala/com/fulcrumgenomics/commons/io/Io.scala
```diff
@@ -226,4 +226,10 @@ trait IoUtil {
...
+    /** Required for 2.12 compatibility. */
+    def underlying: Iterable[A] = this
+  }
```
tfenne: Would it make sense to add an overridable `def close(): Unit = ()`?

tfenne (approved):

PR #61, opened 2020-04-01
Move automatic gzip detection from fgbio to commons.

### LINKED ISSUE #50: Move support for gzipped files from fgbio.Io into commons.Io

sstadick:
I've migrated the gzip auto detection functionality and tests. I left the bgzip outputStream detection in fgbio because it depends on htsjdk which is not in commons.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/commons/io/Io.scala
+++ b/src/main/scala/com/fulcrumgenomics/commons/io/Io.scala
@@ -47,6 +47,7 @@ trait IoUtil {
 
   /** How large a buffer should be used when buffering operations. */
   def bufferSize: Int = 32 * 1024
+  /** The level compression to use when writing compressed output. */
   def compressionLevel: Int = 5
 
   /** Creates a new InputStream to read from the supplied path. Automatically handles gzipped files. */
--- a/src/main/scala/com/fulcrumgenomics/commons/io/Io.scala
+++ b/src/main/scala/com/fulcrumgenomics/commons/io/Io.scala
@@ -47,7 +47,7 @@ trait IoUtil {
 
   /** How large a buffer should be used when buffering operations. */
   def bufferSize: Int = 32 * 1024
-  /** The level compression to use when writing compressed output. */
+  /** The level of compression to use when writing compressed output. */
   def compressionLevel: Int = 5
 
   /** Creates a new InputStream to read from the supplied path. Automatically handles gzipped files. */
--- a/src/main/scala/com/fulcrumgenomics/commons/io/Io.scala
+++ b/src/main/scala/com/fulcrumgenomics/commons/io/Io.scala
@@ -25,6 +25,7 @@ package com.fulcrumgenomics.commons.io
 
 import java.io._
 import java.nio.file.{Files, Path}
+import java.util.zip.{GZIPInputStream, GZIPOutputStream}
 
 import com.fulcrumgenomics.commons.CommonsDef._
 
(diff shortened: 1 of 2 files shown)


sstadick:
Corresponding branch of fgbio: https://github.com/fulcrumgenomics/fgbio/tree/ss_gzip_support_migration

src/main/scala/com/fulcrumgenomics/commons/io/Io.scala
```diff
@@ -46,15 +47,28 @@ trait IoUtil {
 
   /** How large a buffer should be used when buffering operations. */
   def bufferSize: Int = 32 * 1024
+  def compressionLevel: Int = 5
```
tfenne, line 51: Could you just add a docstring for this please?

tfenne (approved):
One minor comment, otherwise LGTM.  For the fgbio PR we'll also need to update the `commons` version to the latest snapshot after this PR is merged otherwise fgbio will temporarily lose this capability.

src/main/scala/com/fulcrumgenomics/commons/io/Io.scala
sstadick: A docstring has been added!

src/main/scala/com/fulcrumgenomics/commons/io/Io.scala
```diff
@@ -46,15 +47,29 @@ trait IoUtil {
...
+    }
   }
 
-  /** Creates a new BufferedReader to read from the supplied path. */
-  def toOutputStream(path: Path) : OutputStream = new BufferedOutputStream(Files.newOutputStream(path), bufferSize)
+  /** Creates a new OutputStream to read from the supplied path. Automatically handles gzipped files. */
+  def toOutputStream(path: Path) : OutputStream = {
+      PathUtil.extensionOf(path) match {
+        case Some(".gz") => new GZIPOutputStream(Files.newOutputStream(path), bufferSize) { this.`def`.setLevel(compressionLevel) }
+        case _           => new BufferedOutputStream(Files.newOutputStream(path), bufferSize)
```
nh13, line 69: @sstadick @tfenne 
1. should we handle block-compressed data here like we do in [fgbio](https://github.com/fulcrumgenomics/fgbio/blob/master/src/main/scala/com/fulcrumgenomics/util/Io.scala#L56)
2.  If not, then we'll need to keep that functionality there when that PR comes about.  As if not, should we remove support for `.bgz` and `.bgzip` above?  Gzip can handle decompression of blocked gzip, but then we have asymmetric support.

src/test/scala/com/fulcrumgenomics/commons/io/IoTest.scala
```diff
@@ -193,4 +194,27 @@ class IoTest extends UnitSpec {
...
+
+  "Io.toOutputStream" should "round trip data to a gzipped file if it ends with .gz" in {
+    val text = "This is a stupid little text fragment for compression. Yay compression!"
+    val in   = Seq(text, text, text)
+    val f = Io.makeTempFile("test", ".gz")
+    Io.writeLines(f, in)
+    val out = Io.readLines(f).toSeq
+    out shouldBe in
+  }
+
```
nh13: newline

src/test/scala/com/fulcrumgenomics/commons/io/IoTest.scala
```diff
@@ -193,4 +194,27 @@ class IoTest extends UnitSpec {
...
+    val reader = new BufferedReader(new InputStreamReader(new GZIPInputStream(new FileInputStream(f.toFile))))
+    val out = Seq(reader.readLine(), reader.readLine(), reader.readLine())
+    out shouldBe in
+    reader.close()
+  }
+
+  "Io.toOutputStream" should "round trip data to a gzipped file if it ends with .gz" in {
+    val text = "This is a stupid little text fragment for compression. Yay compression!"
+    val in   = Seq(text, text, text)
+    val f = Io.makeTempFile("test", ".gz")
```
nh13: align = here and above for practice (looks like wasn't in fgbio)?

src/test/scala/com/fulcrumgenomics/commons/io/IoTest.scala
```diff
@@ -193,4 +194,27 @@ class IoTest extends UnitSpec {
...
+    Io.writeLines(f, in)
+
+    // Manually read it back as gzipped data
+    val reader = new BufferedReader(new InputStreamReader(new GZIPInputStream(new FileInputStream(f.toFile))))
+    val out = Seq(reader.readLine(), reader.readLine(), reader.readLine())
+    out shouldBe in
+    reader.close()
+  }
+
+  "Io.toOutputStream" should "round trip data to a gzipped file if it ends with .gz" in {
```
nh13:
```suggestion
  it should "round trip data to a gzipped file if it ends with .gz" in {
```

nh13 (changes requested):
I think we need to decide on if we want `commons` to handle _reading_ block-compressed data, but not be able to handle `_writing` block-compressed data.  Other than that, only a few minor fixups and then LGTM.

src/main/scala/com/fulcrumgenomics/commons/io/Io.scala
tfenne: In concept it would be nice to have block-gzip support in commons, but _not_ I think if that involves adding `htsjdk` as a dependency for commons, which I suspect it would.  I'm ok with the support being asymmetric.  There are plenty of formats we implement read support for but not write.

sstadick: Either direction ends up being a little awkward. Are we agreeing that asymmetric support is okay for right now?

src/test/scala/com/fulcrumgenomics/commons/io/IoTest.scala
sstadick: Correct, I am a victim of copy paste formatting: https://github.com/fulcrumgenomics/fgbio/blob/b84524e132e5033a0bbe5c6dd661b1b2681c3bdb/src/test/scala/com/fulcrumgenomics/util/IoTest.scala#L40

sstadick: fixed in latest changes

src/test/scala/com/fulcrumgenomics/commons/io/IoTest.scala
sstadick: fixed in latest changes

sstadick:
@tfenne regarding updating the version and cutting a new rerelease, it looks like sbt-release will handle the version bump, and there is nothing to change for this PR?

src/main/scala/com/fulcrumgenomics/commons/io/Io.scala
nh13: Asymmetric is fine. I agree, that i don’t want htsjdk as a dependency in commons.

tfenne:
@sstadick Right - nothing in this PR, but when you make the PR in fgbio to remove this functionality, we'll have to bump the version of commons it depends on.

sstadick:
Sounds good. @nh13, do the small changes look okay to you? If so I can merge this in and get the fgbio side of this in shape.

nh13 (approved):

nh13:
LGTM

PR #62, opened 2020-04-03
Allow IoUtil trait's compressionLevel var to be settable by extending classes

sstadick:
When compressionLevel was a method, it could not be set, which is needed in fgbio on the singleton Io object in FgBioMain. This should fix that now by requiring anything that extends IoUtil to provide a default for the compressionLevel.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/commons/io/Io.scala
+++ b/src/main/scala/com/fulcrumgenomics/commons/io/Io.scala
@@ -35,7 +35,9 @@ import scala.io.Source
 /**
 * Singleton object to provide access to Io utility methods.
 */
-object Io extends IoUtil
+object Io extends IoUtil() {
+  override var compressionLevel: Int = 5
+}
 
 /**
  * Trait that can be mixed in to make an Io utility object, and can be re-used elsewhere.
@@ -46,7 +48,7 @@ trait IoUtil {
   val DevNull: Path = PathUtil.pathTo("/dev/null")
 
   /** The level of compression to use when writing compressed output. */
-  var compressionLevel: Int = 5
+  var compressionLevel: Int
   /** How large a buffer should be used when buffering operations. */
   def bufferSize: Int = 32 * 1024
 
--- a/src/main/scala/com/fulcrumgenomics/commons/io/Io.scala
+++ b/src/main/scala/com/fulcrumgenomics/commons/io/Io.scala
@@ -36,7 +36,7 @@ import scala.io.Source
 * Singleton object to provide access to Io utility methods.
 */
 object Io extends IoUtil() {
-  override var compressionLevel: Int = 5
+  override def compressionLevel: Int = 5
 }
 
 /**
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/commons/io/Io.scala
```diff
@@ -45,10 +47,10 @@ trait IoUtil {
   val StdOut: Path = PathUtil.pathTo("/dev/stdout")
   val DevNull: Path = PathUtil.pathTo("/dev/null")
 
+  /** The level of compression to use when writing compressed output. */
+  var compressionLevel: Int
```
nh13: Alternatively, in fgbio we could create a setter for `compressionLevel`.  I think this is overkill, as there's no logic that this enables that isn't supported by `var`, for example if we need some custom logic or redirection in the getter or setter.  I thought I'd mention it though.
 
```scala
trait Foo extends Io {
  private var _compressionLevel: Int = super.compressionLevel
  def compressionLevel: Int = this._compressionLevel
  def compressionLevel_=(level: Int): Unit = this._compressionLevel = level
}
```

src/test/scala/com/fulcrumgenomics/commons/io/IoTest.scala
```diff
@@ -216,4 +220,9 @@ class IoTest extends UnitSpec {
     val out = Io.readLines(f).toSeq
     out shouldBe in
   }
+  
+  "IoUtil.compressionLevel" should "be settable" in {
+    FakeIo.compressionLevel = 6
+    FakeIo.compressionLevel shouldEqual 6
```
nh13: We use `shouldBe` in most places.  TIL: https://stackoverflow.com/questions/43923312/whats-the-difference-between-shouldbe-vs-shouldequal-in-scala

nh13 (approved):
LGTM, just made a few comments for discussion purposes, but they require no action.

src/test/scala/com/fulcrumgenomics/commons/io/IoTest.scala
sstadick: Noted! I made the change for consistency.

src/main/scala/com/fulcrumgenomics/commons/io/Io.scala
tfenne: I'm not sure I love having the `var` in the trait if I'm honest.   A little reading tells me that scala doesn't allow overriding non-abstract `def`s with `var`s.  But it _does_ allow overriding abstract `def`s.  So the other option would be to just remove the default in the base trait, and force instantiations to provide a value.  Thoughts?

sstadick: Interesting, I'm not sure how I didn't try that! I think that is cleaner as well, and everything seems to work when I do it that way. See newest commit.

sstadick:
@tfenne, how does this look to you? I want to merge this in before https://github.com/fulcrumgenomics/fgbio/pull/575#pullrequestreview-391826913

tfenne:
@sstadick Looks good to me, thanks!

PR #66, opened 2020-10-21
row.get returns Some(" ") instead of returning None when allowMissingColumn flag is set

dsommer:
row.get returns `Some(" ")` instead of returning `None` when `allowMissingColumn` flag is set to true and the field is empty. None is correctly returned when `allowMissingColumn` is not set.
@clintval

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/commons/util/DelimitedDataParser.scala
+++ b/src/main/scala/com/fulcrumgenomics/commons/util/DelimitedDataParser.scala
@@ -74,9 +74,9 @@ class Row private[util] (private val headerIndices: Map[String,Int], private val
     * `allowMissingColumn` is true, then None is returned if the given column name is not present. */
   def get[A](column: String, allowMissingColumn: Boolean = false)(implicit tag: ru.TypeTag[A]): Option[A] =
     (headerIndices.get(column), allowMissingColumn) match {
+      case (Some(idx), _) => get[A](idx)      
       case (None, true)   => None
       case (None, false)  => throw new NoSuchElementException(column)
-      case (Some(idx), _) => get[A](idx)
     }
 }
 
--- a/src/main/scala/com/fulcrumgenomics/commons/util/DelimitedDataParser.scala
+++ b/src/main/scala/com/fulcrumgenomics/commons/util/DelimitedDataParser.scala
@@ -73,7 +73,7 @@ class Row private[util] (private val headerIndices: Map[String,Int], private val
   /** Gets a value from the specified column. If the value is null or empty returns None.  If
     * `allowMissingColumn` is true, then None is returned if the given column name is not present. */
   def get[A](column: String, allowMissingColumn: Boolean = false)(implicit tag: ru.TypeTag[A]): Option[A] =
-    if (allowMissingColumn) headerIndices.get(column).map(apply[A]) else get(headerIndices(column))
+    if( allowMissingColumn && ! headerIndices.contains(column)) None else get(headerIndices(column))
 }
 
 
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/commons/util/DelimitedDataParser.scala
```diff
@@ -73,7 +73,7 @@ class Row private[util] (private val headerIndices: Map[String,Int], private val
   /** Gets a value from the specified column. If the value is null or empty returns None.  If
     * `allowMissingColumn` is true, then None is returned if the given column name is not present. */
```
nh13, line 74: I had to re-read this comment a few times to see what the intended behavior was.  I think will make it clearer, along with your bugfix.
```suggestion
  /** Gets a value from the specified column. If the value is null or empty returns None.  If
    * `allowMissingColumn` is true, then None is also returned if the given column name is not present. */
```

src/main/scala/com/fulcrumgenomics/commons/util/DelimitedDataParser.scala
```diff
@@ -73,7 +73,7 @@ class Row private[util] (private val headerIndices: Map[String,Int], private val
...
   def get[A](column: String, allowMissingColumn: Boolean = false)(implicit tag: ru.TypeTag[A]): Option[A] =
-    if (allowMissingColumn) headerIndices.get(column).map(apply[A]) else get(headerIndices(column))
+    if (allowMissingColumn && ! headerIndices.contains(column)) None else get(headerIndices(column))
```
nh13: How about this? If you go with your implementation, please change `! headerIndices` to `!headerIndices`
```suggestion
    if (allowMissingColumn) headerIndices.get(column).flatMap(get[A]) else get(headerIndices(column))
```

nh13 (approved):

src/test/scala/com/fulcrumgenomics/commons/util/DelimitedDataParserTest.scala
```diff
@@ -210,4 +210,12 @@ class DelimitedDataParserTest extends UnitSpec {
     row.get[Int]("d", allowMissingColumn=true) shouldBe None
     an[Exception] shouldBe thrownBy { row.get[Int]("d", allowMissingColumn=false) }
   }
+
+  it should "test empty fields returning None when allowMissingColumn is true" in {
```
tfenne: the string should describe the intended behavior, not what the test should do.  So in this case perhaps:

```scala
it should "return None for missing values even with allowMissingColumn is true"
```

src/main/scala/com/fulcrumgenomics/commons/util/DelimitedDataParser.scala
tfenne: I find both a bit hard to follow.  I think I would suggest a more verbose approach that's a little easier to read:

```scala
headerIndices.get(column) match {
  case None      => if (allowMissingColumn) None else throw new NoSuchElementException(column)
  case Some(idx) => get[A](idx)
}
```

Either way it would be nice to avoid the double lookup of the column index for the common/positive path.

tfenne (changes requested):
Couple of simple suggestions.

dsommer:
@tfenne I like the verbose match statement solution. Removed `if` statement and added another match case which seems even cleaner. h/t mike
```scala
    (headerIndices.get(column), allowMissingColumn) match {
      case (Some(idx), _) => get[A](idx)
      case (None, true)   => None
      case (None, false)  => throw new NoSuchElementException(column)
    }
```

tfenne (approved):

PR #67, opened 2020-12-16
[bugfix] Sanitize file name now truncates files to avoid filesystem errors on names larger than 255 characters.

jacarey:
Most modern day file systems (ext4, ntfs, xfs, apfs, etc) limit file name size to 255 characters. This PR will ensure that any filename handed back from `sanitizeFileName` will be, at maxiumum, 255 characters.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/commons/io/PathUtil.scala
+++ b/src/main/scala/com/fulcrumgenomics/commons/io/PathUtil.scala
@@ -28,6 +28,9 @@ import java.nio.file.{Path, Paths}
 
 /** Provides utility methods for creating and manipulating Path objects and path-like Strings. */
 object PathUtil {
+  @Deprecated("Use `IllegalCharacters`")
+  def illegalCharacters: String = IllegalCharacters
+
   val MaxFileNameSize: Int = 254
   val IllegalCharacters: String = "[!\"#$%&'()*/:;<=>?@\\^`{|}~] "
 
--- a/src/main/scala/com/fulcrumgenomics/commons/io/PathUtil.scala
+++ b/src/main/scala/com/fulcrumgenomics/commons/io/PathUtil.scala
@@ -28,8 +28,8 @@ import java.nio.file.{Path, Paths}
 
 /** Provides utility methods for creating and manipulating Path objects and path-like Strings. */
 object PathUtil {
-  val maxFileNameSize: Int = 254
-  val illegalCharacters: String = "[!\"#$%&'()*/:;<=>?@\\^`{|}~] "
+  val MaxFileNameSize: Int = 254
+  val IllegalCharacters: String = "[!\"#$%&'()*/:;<=>?@\\^`{|}~] "
 
   /** Resolves a path from a String, and then makes the path absolute. Prefer this to PathUtil.pathTo elsewhere. */
   def pathTo(first: String, more: String*): Path = Paths.get(first, more:_*).toAbsolutePath.normalize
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/commons/io/PathUtil.scala
```diff
@@ -28,12 +28,14 @@ import java.nio.file.{Path, Paths}
 
 /** Provides utility methods for creating and manipulating Path objects and path-like Strings. */
 object PathUtil {
+  val maxFileNameSize: Int = 254
```
nh13: Upper camel case, though `illegalCharacters` is not either :/

src/main/scala/com/fulcrumgenomics/commons/io/PathUtil.scala
```diff
@@ -42,9 +44,12 @@ object PathUtil {
     */
   def sanitizeFileName(fileName: String,
                        illegalCharacters: String = PathUtil.illegalCharacters,
-                       replacement: Option[Char] = Some('_')): String = replacement match {
-    case None    => fileName.filter(c => !illegalCharacters.contains(c))
-    case Some(r) => fileName.map(c => if (illegalCharacters.contains(c)) r else c)
+                       replacement: Option[Char] = Some('_')): String = {
+    val sanitizedFileName = replacement match {
+      case None => fileName.filter(c => !illegalCharacters.contains(c))
+      case Some(r) => fileName.map(c => if (illegalCharacters.contains(c)) r else c)
```
nh13: Don't get rid of the precious `=>` alignment!
```suggestion
      case None    => fileName.filter(c => !illegalCharacters.contains(c))
      case Some(r) => fileName.map(c => if (illegalCharacters.contains(c)) r else c)
```

src/main/scala/com/fulcrumgenomics/commons/io/PathUtil.scala
```diff
@@ -42,9 +44,12 @@ object PathUtil {
     */
   def sanitizeFileName(fileName: String,
                        illegalCharacters: String = PathUtil.illegalCharacters,
-                       replacement: Option[Char] = Some('_')): String = replacement match {
-    case None    => fileName.filter(c => !illegalCharacters.contains(c))
-    case Some(r) => fileName.map(c => if (illegalCharacters.contains(c)) r else c)
+                       replacement: Option[Char] = Some('_')): String = {
```
nh13: How about this, so folks can control it?
```suggestion
                       replacement: Option[Char] = Some('_'),
                       maxFileNameSize: Option[Int] = Some(MaxFileNameSize)): String = {
```

src/test/scala/com/fulcrumgenomics/commons/io/PathUtilTest.scala
```diff
@@ -82,6 +82,21 @@ class PathUtilTest extends UnitSpec {
...
+      "thinkaboutitsincewhywouldyoueverreallyneedafilenamethatisaslongasthisonehonestlyihavenoidea")should be("this" +
+      "isthemostridiculouslylongfilenameandforsomereasonwestillneedtohaveasanitizemethodthatensuresthatnoonecreatesa" +
+      "filenamethatisaslongasthisonewhichiskindofsilywhenyouthinkaboutitsincewhywouldyoueverreallyneedafilenamethat" +
+      "isaslongasthisonehonestlyihavenoi")
+    PathUtil.sanitizeFileName("!!thisisthemostridiculouslylongfilenameandforsomereasonwestillneedtoh" +
+      "aveasanitizemethodthatensuresthatnoonecreatesafilenamethatisaslongasthisonewhichiskindofsilywhenyou" +
+      "thinkaboutitsincewhywouldyoueverreallyneedafilenamethatisaslongasthisonehonestlyihavenoidea",
+      replacement = Some('X')) should be("XXthisisthemostridiculouslylongfilenameandforsomereasonwestillneedtohavea" +
+      "sanitizemethodthatensuresthatnoonecreatesafilenamethatisaslongasthisonewhichiskindofsilywhenyouthinkabout" +
+      "itsincewhywouldyoueverreallyneedafilenamethatisaslongasthisonehonestlyihaven")
```
nh13: Ooff, I think this test would be clearer what it was testing if you constructed the string with `MaxFileNameSize`.  For example:
```scala
PathUtil.sanitizeFileName("*" + (PathUtil.MaxFileNameSize + 10)) shouldBe "*" + PathUtil.MaxFileNameSize
PathUtil.sanitizeFileName("!" + (PathUtil.MaxFileNameSize + 10), replacement = Some('X')) shouldBe "X" * PathUtil.MaxFileNameSize

**Review by nh13 — CHANGES_REQUESTED** [review]

File: src/main/scala/com/fulcrumgenomics/commons/io/PathUtil.scala [hunk 5]
```diff
@@ -28,23 +28,29 @@ import java.nio.file.{Path, Paths}
 
 /** Provides utility methods for creating and manipulating Path objects and path-like Strings. */
 object PathUtil {
-  val illegalCharacters: String = "[!\"#$%&'()*/:;<=>?@\\^`{|}~] "
+  val MaxFileNameSize: Int = 254
+  val IllegalCharacters: String = "[!\"#$%&'()*/:;<=>?@\\^`{|}~] "
```
nh13, line 35: Do you want to have a method `def illegalCharacters: String = IllegalCharacters` that's deprecated so we don't break the API?

nh13 (approved):
One last suggestion, then it's ready to go

PR #68, opened 2021-01-03
Use thread-safe DateTimeFormatter in log emissions

clintval:
Unfortunately `SimpleDateFormat` is not thread-safe and must be synchronized. Instead of synchronizing access to the class, I replaced it with the Java 8 `DateTimeFormatter` which is thread-safe by default. I believe this will make synchronous access to these loggers possible. Snippet from the unit tests show no functional change:

```console
[2021/01/03 17:19:15 | DelimitedDataParser | Error] Line has incorrect number of fields. Header length is 3, row length is 4
[2021/01/03 17:19:15 | DelimitedDataParser | Error] Headers: one,two,three
[2021/01/03 17:19:15 | DelimitedDataParser | Error] Line: one,two,three,four
[2021/01/03 17:19:15 | DelimitedDataParser | Error] Line has incorrect number of fields. Header length is 3, row length is 2
[2021/01/03 17:19:15 | DelimitedDataParser | Error] Headers: one,two,three
[2021/01/03 17:19:15 | DelimitedDataParser | Error] Line: one,two
```

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/commons/util/Logger.scala
+++ b/src/main/scala/com/fulcrumgenomics/commons/util/Logger.scala
@@ -24,8 +24,8 @@
 package com.fulcrumgenomics.commons.util
 
 import java.io.{PrintStream, PrintWriter, StringWriter}
-import java.text.SimpleDateFormat
-import java.util.Date
+import java.time.LocalDateTime
+import java.time.format.DateTimeFormatter
 
 import scala.collection.compat._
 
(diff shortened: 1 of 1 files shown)


tfenne:
@clintval I don't think `Logger` was ever really intended to be thread-safe, but looking at the impementation it is that way more or less accidentally except for the use of `SimpleDateFormat`.  I see no harm in merging this.  @nh13 ?

clintval:
Eye-ball static analysis suggests it will become thread-safe after this PR...

clintval:
Should I update the documentation of the class to note our intent is a thread-safe implementation?

nh13:
Lgtm

tfenne:
@clintval yes please on updating the docs and then we'll merge this.

clintval:
All set!

PR #71, opened 2021-07-30
Make BiMap a better all around Scala Map collection

clintval:
I have a use case for `BiMap` and need it to act more like a real Scala collection.

This compiles on 2.13.

The class builder is compatible with both 2.12 and 2.13.

However, I don't know how to make backward compat for these Scala 2.13 traits:

- `MapOps`
- `StrictOptimizedMapOps`
- `StrictOptimizedIterableOps`

In the meantime, I can compile this directly into my own 2.13 project. Feel free to take this PR or close it. It might be nice to have as an upgrade if you ever drop 2.12.


clintval:
I ended up putting my work here: https://github.com/clintval/bimap

PR #72, opened 2021-09-10
[feature] a least recently used cache collection

nh13:


src/main/scala/com/fulcrumgenomics/commons/collection/LeastRecentlyUsedCache.scala
```diff
@@ -0,0 +1,71 @@
...
+ */
+
+package com.fulcrumgenomics.commons.collection
+
+/** A cache that stores up to a maximum number of entries, that when reached,
+  * evicts the least recently used (by get or put) key. */
+class LeastRecentlyUsedCache[Key, Value](maxEntries: Int) extends Iterable[(Key, Value)] {
+
+  private val map = scala.collection.mutable.LinkedHashMap[Key, Value]()
+
```
tfenne, line 35: require maxEntires >= 1?

src/main/scala/com/fulcrumgenomics/commons/collection/LeastRecentlyUsedCache.scala
```diff
@@ -0,0 +1,71 @@
...
+  /** Gets value for the given key, or None if it doesn't exist.  Will update the key to
+    * be the most recently used. */
+  def get(key: Key): Option[Value] = map.synchronized {
+    val result = map.get(key)
+    // remove and add the key/value if found, to move it to the most recently used
+    result.foreach { value =>
+      map.remove(key)
+      map.put(key, value)
+    }
+    result
```
tfenne:
```suggestion
    val result = map.remove(key)
    // re-insert if found, to move it to the most recently used
    result.foreach { value => map.put(key, value) }
    result
```

src/main/scala/com/fulcrumgenomics/commons/collection/LeastRecentlyUsedCache.scala
```diff
@@ -0,0 +1,71 @@
...
+    // remove the least recently used element if needed
+    // IMPORTANT: must check the size **after** removing the key, since if the key is present, we're overwriting it and
+    // so the # of entries wont change.
+    if (map.size == maxEntries) { // this implies the key was not present in the cache
+      map.headOption.foreach { case (k, _) => map.remove(k) }
+    }
+    // add the key/value as the most recently used
+    map.put(key, value)
+    // return the removed value
+    previous
```
tfenne:
```suggestion
    val previous = map.put(key, value)

    if (map.size > maxEntries) {
      map.keysIterator.take(map.size - maxEntries).foreach(map.remove)
    }

   previous
```

src/main/scala/com/fulcrumgenomics/commons/collection/LeastRecentlyUsedCache.scala
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  /** Removes the value with the given key from the cache. */
+  def remove(key: Key): Option[Value] = map.remove(key)
```
tfenne: Needs to be synchronized.

src/main/scala/com/fulcrumgenomics/commons/collection/LeastRecentlyUsedCache.scala
```diff
@@ -0,0 +1,71 @@
...
+
+  /** An iterator over key-value pairs from least to most recently used. */
+  override def iterator: Iterator[(Key, Value)] = map.iterator
+
+  /** The number elements in the cache. */
+  override def size: Int = map.size
```
tfenne: Needs to be synchronized.

src/main/scala/com/fulcrumgenomics/commons/collection/LeastRecentlyUsedCache.scala
```diff
@@ -0,0 +1,71 @@
...
+    map.put(key, value)
+    // return the removed value
+    previous
+  }
+
+  /** Removes the value with the given key from the cache. */
+  def remove(key: Key): Option[Value] = map.remove(key)
+
+  /** An iterator over key-value pairs from least to most recently used. */
+  override def iterator: Iterator[(Key, Value)] = map.iterator
```
tfenne: I'm not sure how safe this is.  Maybe at least document that this should not be called or the returned iterator used  concurrently with any of the methods that modify the cache?

tfenne (changes requested):

src/main/scala/com/fulcrumgenomics/commons/collection/LeastRecentlyUsedCache.scala
nh13: This doesn't work as the `map.put` doesn't re-insert an existing key, thereby not updating it's insertion order in the underlying `LinkedHashMap`

tfenne: Ah, I see.

src/main/scala/com/fulcrumgenomics/commons/collection/LeastRecentlyUsedCache.scala
```diff
@@ -0,0 +1,76 @@
...
+  /** Adds the value for the given key.  Will update the key to be the most recently used. Returns
+    * the value that was removed, or None if the key was not in the cache. */
+  def put(key: Key, value: Value): Option[Value] = map.synchronized {
+    // If the key exists in the underyling map, it needs to be re-inserted to update the insertion order used in
+    // LinkedHashMap.  If the key is not present, previous is None.
+    val previous = map.remove(key)
+    // Remove the least recently used element if needed
+    // IMPORTANT: must check the size **after** removing the key, since if the key is present, we're overwriting it and
+    // so the # of entries wont change.
+    if (map.size == maxEntries) { // this implies the key was not present in the cache
```
tfenne, line 54: I think I would still tend to make this `while (map.size >= maxEntries)` just to be on the safe side - e.g. if someone grabbed the iterator and inserted things through it, but it's not critical since it shouldn't ever go over.

tfenne (approved):

src/main/scala/com/fulcrumgenomics/commons/collection/LeastRecentlyUsedCache.scala
nh13: I changed the iterator so they can't do that. I was proud of myself for this, so I'm going to keep it.

PR #73, opened 2021-11-01
SimpleCounter iterators are in insertion order

nh13:


nh13:
@tfenne Happy to add a unit test if you agree.

tfenne:
@nh13 I think this is nice (at least gives _some_ deterministic order) but just would want to make sure it's not massively slower to use a LinkedHashMap - hopefully once the value is inserted the overhead is zero?

Also, I can haz a `def keys: Iterator[T]`?

tfenne (approved):

PR #74, opened 2021-12-19
Fix AsyncIterator race condition when first element in source raises

clintval:
This PR fixes a race condition with the exception-raising mechanics of `AsyncIterator`.

The bug occurs when the head (or the first few) elements of the source iterator raise exceptions. The prior mechanics of `AsyncIterator` might run the check-and-raise call in the `hasNext()` block _**before**_ any of the first elements are actually computed. This means that exceptions can be swallowed and your iterator will be silently truncated to 0-length.

A quick example of how the race condition will create missing data:

```scala
 def raiseOnTen(num: Int): Int = if (num == 10) throw new IllegalArgumentException else num
 new AsyncIterator(Seq(10, 11, 12, 13).iterator.map(raiseOnTen)).start().toSeq == Seq.empty
```

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/commons/async/AsyncIterator.scala
+++ b/src/main/scala/com/fulcrumgenomics/commons/async/AsyncIterator.scala
@@ -25,8 +25,6 @@
 
 package com.fulcrumgenomics.commons.async
 
-import com.fulcrumgenomics.commons.CommonsDef.yieldAndThen
-
 import java.util.concurrent.{ArrayBlockingQueue, BlockingQueue, LinkedBlockingDeque, TimeUnit}
 
 object AsyncIterator {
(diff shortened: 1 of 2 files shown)


clintval:
Tests had passed on my local... Looks like there are some more race conditions....

nh13:
@clintval let me know if you want me to take a look at the code, the additional race condition, or otherwise

clintval:
If you're volunteering, that would kind! It's not easy code to debug.

I _believe_ I fixed 1 race condition but I am certain there are also others. I might do some OSS contributions over break and could get to this. I actually think I know how to fix this one but no promises.

nh13:
Why don't you give it a shot, then I will.

clintval:
@nh13 I'm certain I improved the implementation. I _think_ I fixed up to 2 ways the race condition can manifest. But unfortunately I do not know how to prove there are no more race conditions. The race condition occurs when there are exceptions raised _within_ the input source iterator which will short-circuit the iterator and (with the prior implementation) go silent.

clintval:
If you remove my fix in the class and run the new test you will see it often fails:

```scala
❯ testOnly com.fulcrumgenomics.commons.async.AsyncIteratorTest
[info] - should correctly propagate an exception that originates from within the source iterator *** FAILED ***
[info]   Expected exception java.lang.IllegalArgumentException to be thrown, but no exception was thrown (AsyncIteratorTest.scala:51)
```

clintval:
I thought about it for a few more hours and realized I made the race condition occur more infrequently but I did not actually solve it. Now I believe I have solved it (famous last words).

nh13 (approved):

nh13:
@clintval thank-you thank-you!!!

PR #75, opened 2022-02-19
Make sure IntelDeflator is shown as unsupported on non-x86 Apple machines e.g. M1x.

tfenne:
@clintval would you mind taking a look at this too please?

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/commons/util/SystemUtil.scala
+++ b/src/main/scala/com/fulcrumgenomics/commons/util/SystemUtil.scala
@@ -58,7 +58,7 @@ object SystemUtil {
   /** True if the current system might support the Intel Inflater and Deflater, false otherwise. */
   lazy val IntelCompressionLibrarySupported: Boolean = {
     if (!isLinux && !isMac) false
-    else if (isOsArch(name = "ppc64le")) false
+    else if (isMac && !isOsArch("x86")) false
     else if (isMac && isOsVersion("10\\.14\\..+".r)) false // FIXME: https://github.com/Intel-HLS/GKL/issues/101
     else true
   }


src/main/scala/com/fulcrumgenomics/commons/util/SystemUtil.scala
```diff
@@ -58,7 +58,7 @@ object SystemUtil {
   /** True if the current system might support the Intel Inflater and Deflater, false otherwise. */
   lazy val IntelCompressionLibrarySupported: Boolean = {
     if (!isLinux && !isMac) false
-    else if (isOsArch(name = "ppc64le")) false
+    else if (isMac && !isOsArch("x86")) false
```
clintval, line 61: On my 2.3 GHz 8-Core Intel Core i9 I get `"x86_64"` for this value, and on my Apple M1 Max I get `"aarch64"`.

I think because the function uses `.contains()` and not exact match, it's fine, but this would be more clear?

```suggestion
    else if (isMac && isOsArch("aarch64")) false // Apple silicon is not supported by Intel GKL
```

I think you can delete line 62 since Intel GKL [should have a patch in 0.8.7](https://github.com/Intel-HLS/GKL/issues/101#issuecomment-648342314) and we are now on [0.8.8](https://github.com/fulcrumgenomics/fgbio/blob/9054893b02f5d5e1cbe0322daa58b634bbd00447/build.sbt#L136).

clintval (approved):

src/main/scala/com/fulcrumgenomics/commons/util/SystemUtil.scala
clintval: Anecdotally, I am noticing the Java inflaters/deflaters are getting faster and I still don't have any data to suggest recent Intel inflater/deflaters are actually any better; in fact, we have some data to suggest they are getting slower!

src/main/scala/com/fulcrumgenomics/commons/util/SystemUtil.scala
```diff
@@ -58,7 +58,7 @@ object SystemUtil {
...
     else if (isMac && isOsVersion("10\\.14\\..+".r)) false // FIXME: https://github.com/Intel-HLS/GKL/issues/101
```
clintval, line 62: I don't know if anyone is living in the past so it could be fine to leave this here. I'm up to `11.6.2` now.

src/main/scala/com/fulcrumgenomics/commons/util/SystemUtil.scala
tfenne: I think I prefer to just have that `not contains x86` as that really is the constraint - either we have x86 CPUs and the intel inflater/deflator can work, or we don't and they don't.

Interesting on the performance front.  I tend to think that the intel ones are more hassle than they're worth but @nh13 still seems to like them.

PR #76, opened 2022-02-26
A parallel iterator implementation.

tfenne:
This code was extracted and lightly modified from https://github.com/fulcrumgenomics/fgbio/pull/790, with tests added.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/commons/CommonsDef.scala
+++ b/src/main/scala/com/fulcrumgenomics/commons/CommonsDef.scala
@@ -25,15 +25,13 @@
 package com.fulcrumgenomics.commons
 
 import java.io.Closeable
-
-import com.fulcrumgenomics.commons.collection.BetterBufferedIterator
+import com.fulcrumgenomics.commons.collection.{BetterBufferedIterator, ParIterator}
 import com.fulcrumgenomics.commons.util.Logger
 
 import scala.collection.compat._
 import scala.collection.parallel.immutable
 import scala.collection.parallel.{ForkJoinTaskSupport, ParIterable, TaskSupport}
 import java.util.concurrent.ForkJoinPool
-
 import scala.language.implicitConversions
 import scala.util.{Failure, Success, Try}
 
@@ -371,6 +369,13 @@ trait CommonsDef extends Compat {
     /** Creates a parallel collection with the desired level of parallelism and FIFO semantics. */
     def parWith(parallelism: Int, fifo: Boolean = true): P = {
       parWith(new ForkJoinPool(parallelism, ForkJoinPool.defaultForkJoinWorkerThreadFactory, null, fifo))
+    }
+  }
+
+  /** Implicit class that allows generation of ParIterators from Iterators with convenience methods. */
+  implicit class ParIteratorSupport[A](iterator: Iterator[A]) {
+    def parWith(threads: Int, chunkSize: Int = 2048, chunkBuffer: Int = 2): ParIterator[A] = {
+      ParIterator(this.iterator, chunkSize=chunkSize, threads=threads, if (chunkBuffer < 1) None else Some(chunkBuffer))
     }
   }
 
--- a/src/main/scala/com/fulcrumgenomics/commons/CommonsDef.scala
+++ b/src/main/scala/com/fulcrumgenomics/commons/CommonsDef.scala
@@ -374,8 +374,18 @@ trait CommonsDef extends Compat {
 
   /** Implicit class that allows generation of ParIterators from Iterators with convenience methods. */
   implicit class ParIteratorSupport[A](iterator: Iterator[A]) {
-    def parWith(threads: Int, chunkSize: Int = 2048, chunkBuffer: Int = 2): ParIterator[A] = {
-      ParIterator(this.iterator, chunkSize=chunkSize, threads=threads, if (chunkBuffer < 1) None else Some(chunkBuffer))
+    /**
+      * Creates a [[ParIterator]]; see documentation for that class for detailed usage.
+      *
+      * @param threads the number of threads to use in parallel transform operations on the iterator
+      * @param chunkSize the size of chunks to collect and perform parallel operations on
+      * @param chunkBuffer if > 0 use an [[com.fulcrumgenomics.commons.async.AsyncIterator]] to accumulate/cache
+      *                    `chunkBuffer` incoming chunks ready for parallel processing
+      */
+    def parWith(threads: Int,
+                chunkSize: Int = ParIterator.DefaultChunkSize,
+                chunkBuffer: Int = ParIterator.DefaultChunkBuffer): ParIterator[A] = {
+      ParIterator(this.iterator, threads=threads, chunkSize=chunkSize, chunkBuffer=chunkBuffer)
     }
   }
 
--- a/src/main/scala/com/fulcrumgenomics/commons/collection/ParIterator.scala
+++ b/src/main/scala/com/fulcrumgenomics/commons/collection/ParIterator.scala
@@ -102,24 +102,22 @@ object ParIterator {
 class ParIterator[A] private (private val iter: Iterator[Seq[A]],
                               val chunkSize: Int = 1024,
                               private val taskSupport: TaskSupport) extends AbstractIterator[A] {
-  private var currentChunk: Seq[A] = if (iter.hasNext) iter.next() else Seq.empty
+  private var currentChunk: Iterator[A] = if (iter.hasNext) iter.next().iterator else Iterator.empty
 
   /** True if there are more items available, false otherwise. */
   override def hasNext: Boolean = {
-    while (currentChunk.isEmpty && iter.hasNext) currentChunk = iter.next()
+    while (currentChunk.isEmpty && iter.hasNext) currentChunk = iter.next().iterator
     currentChunk.nonEmpty
   }
 
   /** Retrieve the next item.  Will throw an exception if there are no more items. */
   override def next(): A = {
     if (!hasNext) throw new NoSuchElementException
-    val v = currentChunk.head
-    currentChunk = currentChunk.tail
-    v
+    currentChunk.next()
   }
 
   /** Returns an iterator over all remaining chunks including the current chunk. */
-  private def chunks: Iterator[Seq[A]] = Iterator(currentChunk) ++ iter
+  private def chunks: Iterator[Seq[A]] = Iterator(currentChunk.to(LazyList)) ++ iter
 
   /** Wraps an async iterator around this parallel iterator to draw items through the iterator. This should only
     * be invoked after any other transforming operations such as map/filter/etc.
--- a/src/main/scala/com/fulcrumgenomics/commons/collection/ParIterator.scala
+++ b/src/main/scala/com/fulcrumgenomics/commons/collection/ParIterator.scala
@@ -125,7 +125,13 @@ class ParIterator[A] private (private val iter: Iterator[Seq[A]],
     * @param cache how many elements to accumulate in the async iterator's cache - for best memory usage this
     *              should be a multiple of `chunkSize`.
     */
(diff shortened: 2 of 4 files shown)


src/main/scala/com/fulcrumgenomics/commons/CommonsDef.scala
```diff
@@ -374,6 +372,13 @@ trait CommonsDef extends Compat {
     }
   }
 
+  /** Implicit class that allows generation of ParIterators from Iterators with convenience methods. */
+  implicit class ParIteratorSupport[A](iterator: Iterator[A]) {
+    def parWith(threads: Int, chunkSize: Int = 2048, chunkBuffer: Int = 2): ParIterator[A] = {
```
tfenne: I thought about adding `parWith(pool)` and `parWith(support)` here too, but it's complicated by the fact that we can't have multiple methods with the same name and default parameter values.  Since we hardly ever use those variants I've left them off for now.  We could solve this be either:

1. Not having defaults on the other methods
2. Giving them different names (e.g. parWithPool(...))
3. Having multiple implicit classes all with a single `parWith` method each

tfenne:
The PR build is failing under scala 2.12 - should we also desupport 2.12 for commons like we did for fgbio?

src/main/scala/com/fulcrumgenomics/commons/CommonsDef.scala
nh13: I definitely use `parWith(pool)` but only because I've chunked it up and don't want to create a new thread pool for each chunk, so this solves that!  🤣

nh13: I don't know what `chunkSize` and `chunkBuffer` do without looking at the code, can you add param docs or link to `ParIterator`?

src/main/scala/com/fulcrumgenomics/commons/collection/ParIterator.scala
```diff
@@ -0,0 +1,147 @@
...
+
+import com.fulcrumgenomics.commons.CommonsDef._
+import com.fulcrumgenomics.commons.async.AsyncIterator
+
+import java.util.concurrent.ForkJoinPool
+import scala.collection.AbstractIterator
+import scala.collection.parallel.{ForkJoinTaskSupport, TaskSupport}
+
+/**
+  * Companion object to help with manufacturing ParIterators.
```
nh13:
```suggestion
  * Methods to help with manufacturing [[ParIterators]].
```

src/main/scala/com/fulcrumgenomics/commons/collection/ParIterator.scala
```diff
@@ -0,0 +1,147 @@
...
+  */
+object ParIterator {
+  /**
+    * Constructs a ParIterator from the underlying iterator, that will perform operations in parallel on chunks
```
nh13: Helps with java doc here and throughout.
```suggestion
    * Constructs a [[ParIterator]] from the underlying iterator that will perform operations in parallel on chunks
```

src/main/scala/com/fulcrumgenomics/commons/collection/ParIterator.scala
```diff
@@ -0,0 +1,147 @@
...
+    * of `chunkSize` using `threads` threads.
+    *
+    * @param iter the underlying iterator to parallelize over
+    * @param chunkSize the number of elements to chunk together for parallel processing
+    * @param threads the number of threads to use for parallel computations like `map()`
+    * @param chunkBuffer if Some(n) use an [[AsyncIterator]] to pre-buffer n chunks from the input
```
nh13:
```suggestion
    * @param chunkBuffer if Some(n) use an [[AsyncIterator]] to pre-buffer `n` chunks from the input
```

src/main/scala/com/fulcrumgenomics/commons/collection/ParIterator.scala
```diff
@@ -0,0 +1,147 @@
...
+    *                    read for parallel computation
+    */
+  def apply[A](iter: Iterator[A], chunkSize: Int = 1024, threads: Int, chunkBuffer: Option[Int] = Some(2)): ParIterator[A] = {
```
nh13: nit: move threads to before chunkSize in the parameter order, it has no default.

src/main/scala/com/fulcrumgenomics/commons/collection/ParIterator.scala
```diff
@@ -0,0 +1,147 @@
...
+  * ParIterator(iter, chunkSize=4096, threads=8)
+  *   .filter(_.contains("foo"))
+  *   .map(_.toUpperCase)
+  *   .map(_.reverse)
+  *   .toAsync(cache=4096 * 8)
+  *   .foreach(println)
+  * }}}
+  */
+class ParIterator[A] private (private val iter: Iterator[Seq[A]],
+                              val chunkSize: Int = 1024,
```
nh13, line 103: defaults could be moved into the companion object.  Looks like this default is used in a few places.

src/main/scala/com/fulcrumgenomics/commons/collection/ParIterator.scala
```diff
@@ -0,0 +1,147 @@
...
+    val v = currentChunk.head
+    currentChunk = currentChunk.tail
+    v
+  }
+
+  /** Returns an iterator over all remaining chunks including the current chunk. */
+  private def chunks: Iterator[Seq[A]] = Iterator(currentChunk) ++ iter
+
+  /** Wraps an async iterator around this parallel iterator to draw items through the iterator. This should only
+    * be invoked after and other transforming operations such as map/filter/etc.
```
nh13: `after and other` doesn't track for me

src/main/scala/com/fulcrumgenomics/commons/collection/ParIterator.scala
```diff
@@ -0,0 +1,147 @@
...
+                              val chunkSize: Int = 1024,
+                              private val taskSupport: TaskSupport) extends AbstractIterator[A] {
+  private var currentChunk: Seq[A] = if (iter.hasNext) iter.next() else Seq.empty
+
+  /** True if there are more items available, false otherwise. */
+  override def hasNext: Boolean = currentChunk.nonEmpty || iter.hasNext
+
+  /** Retrieve the next item.  Will throw an exception if there are no more items. */
+  override def next(): A = {
+    while (currentChunk.isEmpty && iter.hasNext) currentChunk = iter.next()
```
nh13: why `while` and not `if`?  Do we expect the iterator to return empty groups?

src/main/scala/com/fulcrumgenomics/commons/collection/ParIterator.scala
```diff
@@ -0,0 +1,147 @@
...
+    val v = currentChunk.head
+    currentChunk = currentChunk.tail
+    v
```
nh13: Any benefit to using a linkedlist (list)?  Or should `currentChunk` be an iterator so you can do avoid the above and just call `currentChunk.next()`.

src/main/scala/com/fulcrumgenomics/commons/collection/ParIterator.scala
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  /** Returns an iterator over all remaining chunks including the current chunk. */
+  private def chunks: Iterator[Seq[A]] = Iterator(currentChunk) ++ iter
+
+  /** Wraps an async iterator around this parallel iterator to draw items through the iterator. This should only
+    * be invoked after and other transforming operations such as map/filter/etc.
+    *
+    * @param cache how many elements to accumulate in the async iterator's cache - for best memory usage this
+    *              should be a multiple of chunkSize.
```
nh13:
```suggestion
    *              should be a multiple of `chunkSize`.
```

src/main/scala/com/fulcrumgenomics/commons/collection/ParIterator.scala
```diff
@@ -0,0 +1,147 @@
...
+    */
+  def toAsync(cache: Int = 4 * chunkSize): AsyncIterator[A] = AsyncIterator(this, Some(cache))
```
nh13: part of me wants the async iterator to be able to use the threads in the pool.  Not sure how feasible that is given AysncIterator is a daemon thread/runnable

src/test/scala/com/fulcrumgenomics/commons/collection/ParIteratorTest.scala
```diff
@@ -0,0 +1,47 @@
...
+
+import com.fulcrumgenomics.commons.util.UnitSpec
+
+class ParIteratorTest extends UnitSpec {
+  "ParIterator" should "map the same as a non-parallel map" in {
+    val xs  = Range.inclusive(1, 100)
+    val par = ParIterator(xs.iterator, chunkSize=8, threads=4, chunkBuffer=Some(3))
+    val out = par.map(_ + 1).map(_ * 2).toAsync().toIndexedSeq
+    val exp = xs.map(_ + 1).map(_ * 2)
+    out shouldBe exp
```
nh13: How do you feel about this verbosity?
```suggestion
    out contain theSameElementsInOrderAs exp
```

src/test/scala/com/fulcrumgenomics/commons/collection/ParIteratorTest.scala
```diff
@@ -0,0 +1,47 @@
...
+  it should "allow for a more complicated chain of operations" in {
+    val xs  = Range.inclusive(1, 100)
+    val par = ParIterator(xs.iterator, chunkSize=8, threads=4, chunkBuffer=Some(3))
+    val out = par.filter(_ % 3 == 0).flatMap(x => Seq(x, x)).filterNot(_ < 50).toAsync().toIndexedSeq
+    val exp =  xs.filter(_ % 3 == 0).flatMap(x => Seq(x, x)).filterNot(_ < 50)
+    out shouldBe exp
+  }
+
+  Seq(None, Some(2)).foreach { chunkBuffer =>
+    it should f"propogate errors when chunkBuffering is ${chunkBuffer.isDefined}" in {
```
nh13:
```suggestion
    it should f"propagate errors when chunkBuffer is ${chunkBuffer.isDefined}" in {
```

nh13 (approved):

nh13:
Lets remove support for 2.12.  It's been a while.

src/main/scala/com/fulcrumgenomics/commons/collection/ParIterator.scala
tfenne: It's possible given the way that the `filter`/`filterNot`/`flatMap` mehods are implemented, the chunks are filtered/mapped on the fly and could result in empty-chunks post filtering.

src/main/scala/com/fulcrumgenomics/commons/collection/ParIterator.scala
tfenne: It's not currently a linked list - the way `grouped` works it returns an `ArraySeq` I think.  I tried all sorts of things here including creating a custom mutable Slice type that just wrapped another array/collection and bumped a `start` pointer and none of it made a difference to performance.

Making it an iterator makes the implementation clunky as then all the methods below have to transform one iterator and many Seqs.

src/main/scala/com/fulcrumgenomics/commons/collection/ParIterator.scala
tfenne: I think that would be really hard.

src/main/scala/com/fulcrumgenomics/commons/CommonsDef.scala
tfenne: Done.

src/main/scala/com/fulcrumgenomics/commons/collection/ParIterator.scala
tfenne: Done

PR #88, opened 2025-03-20
feat: expose the header in Row

### LINKED ISSUE #87: `Row` should expose the headers so we don't have to pass it around to folks who use row
See: https://github.com/fulcrumgenomics/fgbio/pull/826#discussion_r2006105417

nh13:
Fixes #87 

This is useful when `Row` is being passed around.  I kept it as `Iterable` in a `def` so that we don't consume more memory.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/commons/util/DelimitedDataParser.scala
+++ b/src/main/scala/com/fulcrumgenomics/commons/util/DelimitedDataParser.scala
@@ -78,6 +78,9 @@ class Row private[util] (private val headerIndices: Map[String,Int], private val
       case (None, true)   => None
       case (None, false)  => throw new NoSuchElementException(column)
     }
+
+  /** Gets the header key (column name) for this row */
+  def header: Iterable[String] = this.headerIndices.keys
 }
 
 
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/commons/util/DelimitedDataParser.scala
```diff
@@ -78,6 +78,9 @@ class Row private[util] (private val headerIndices: Map[String,Int], private val
       case (None, true)   => None
       case (None, false)  => throw new NoSuchElementException(column)
     }
+
+  /** Gets the header key (column name) for this row */
```
nh13:
```suggestion
  /** Gets the header keys (column names) for this row */
```

clintval (approved):
I like it, thanks!

src/test/scala/com/fulcrumgenomics/commons/util/DelimitedDataParserTest.scala
```diff
@@ -218,4 +218,10 @@ class DelimitedDataParserTest extends UnitSpec {
     row.get[String]("c", allowMissingColumn=true) shouldBe None
   }
 
+  "DelimitedDataParser.header" should "return the column header" in {
```
clintval:
```suggestion
  "Row.header" should "return the header for all columns" in {
```

PR #92, opened 2025-10-03
Ensure file toRealPath() does not raise for file descriptors

clintval:
The implementation of [`.toRealPath()`](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Path.html#toRealPath-java.nio.file.LinkOption...-) is platform-dependent and I discovered will raise an exception if called on a file descriptor on Unix platforms (like Linux) but not on Mac. This exception is not caught and halts the program. For example:

```console
Exception in thread "main" java.nio.file.NoSuchFileException: /dev/fd/63
    at java.base/sun.nio.fs.UnixException.translateToIOException(UnixException.java:92)
    at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:106)
    at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:111)
    at java.base/sun.nio.fs.UnixPath.toRealPath(UnixPath.java:886)
    at com.fulcrumgenomics.commons.io.IoUtil.toInputStream(Io.scala:65)
    at com.fulcrumgenomics.commons.io.IoUtil.toInputStream$(Io.scala:58)
    at com.fulcrumgenomics.util.Io.toInputStream(Io.scala:37)
    at com.fulcrumgenomics.commons.io.IoUtil.toSource(Io.scala:96)
    at com.fulcrumgenomics.commons.io.IoUtil.toSource$(Io.scala:95)
    at com.fulcrumgenomics.util.Io.toSource(Io.scala:37)
    at com.fulcrumgenomics.commons.io.IoUtil.readLines(Io.scala:213)
    at com.fulcrumgenomics.commons.io.IoUtil.readLines$(Io.scala:213)
    at com.fulcrumgenomics.util.Io.readLines(Io.scala:37)
```

When trying to read an input file with syntax like:

```bash
tool --input <( ... )
```

To solve this issue, we classify all paths that raise an exception when `.toRealPath()` is called as special files and open them for reading using a non-buffered input stream. I also confirmed that buffered input streaming does not work on named pipes and also fails on file descriptors too. I updated the line comments to reflect that.

I tested this on both Mac and Linux platforms and this PR fixes my issues.

### DIFF
--- a/src/main/scala/com/fulcrumgenomics/commons/io/Io.scala
+++ b/src/main/scala/com/fulcrumgenomics/commons/io/Io.scala
@@ -27,11 +27,11 @@ import java.io._
 import java.nio.file.attribute.BasicFileAttributes
 import java.nio.file.{Files, Path}
 import java.util.zip.{GZIPInputStream, GZIPOutputStream}
-
 import com.fulcrumgenomics.commons.CommonsDef._
 
 import scala.collection.compat._
 import scala.io.Source
+import scala.util.{Failure, Success, Try}
 
 /**
 * Singleton object to provide access to Io utility methods.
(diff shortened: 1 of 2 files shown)


src/main/scala/com/fulcrumgenomics/commons/io/Io.scala
```diff
@@ -60,19 +60,19 @@ trait IoUtil {
...
+        // Check if this path is to a special file or not. Although the behavior of toRealPath() is platform-dependent,
+        // exceptions may raise if the path is a file descriptor so we want to return true in those cases. Later we
+        // choose to buffer reading from regular files only since buffered reading from named pipes and file descriptors
+        // can go poorly due to the BufferedInputStream trying to call .available() on the stream which subsequently
+        // calls .seek() which is illegal to do on pipes and file descriptors.
+        val isSpecialFile = Try(Files.readAttributes(path.toRealPath(), classOf[BasicFileAttributes]))
+          .map(attrs => !attrs.isRegularFile && attrs.isOther)
+          .getOrElse(true)
+
+        if (isSpecialFile) stream else new BufferedInputStream(stream, bufferSize)
```
nh13, line 75: Sips tea: https://github.com/fulcrumgenomics/fgpyo/blob/39a7f58bcbe6eac7bd9ee667c889a8e558b8b237/fgpyo/io/__init__.py#L75-L77

nh13 (approved):

PR #96, opened 2025-11-21
Ensure Io.assertReadable works on file descriptors too

clintval:
The function `Files.isReadable` does not work on file descriptors, unfortunately. So for any path that looks like a file descriptor, we should skip the `Files.isReadable` check.  I couldn't think or find of a better way of testing readability of a file descriptor without causing a side effect (such as by trying to open it).


nh13 (changes requested):
~~Can you add a unit test?~~

nh13 (approved):
NVMD, I see you're full comment now

clintval:
> ~Can you add a unit test?~

I thought about this. It would be difficult to do in a cross-platform way since there is no way to create a file descriptor from within Java/Scala and I would have to create one using the outside environment and then have an integration test of some kind. I suppose I could send in a "fake" file descriptor path that was mocked, but that wouldn't really test what I'm trying to test since my patch only does path prefix matching, and not actually inspecting the filesystem object itself.

PR #99, opened 2026-02-06
feat: ensure toInputStream works on file descriptors

clintval:
This is the same issue I hit with `Io.assertReadable` but this time on `Io.toInputStream`:

- https://github.com/fulcrumgenomics/commons/pull/96

We need to guard for file descriptors before calling `Files.isSameFile` which raises exceptions on special file descriptors. I have tested that this works on a system with file descriptors.


nh13 (approved):
@clintval I searched for `isSameFile` and it looks like this is the last one!

clintval:
> @clintval I searched for `isSameFile` and it looks like this is the last one!

Thank goodness 😅  I feel like this is the third time I've hit an issue like this!

